PR #2, opened 2024-02-20
Add language definition and sample

tbennun:
Using Lark


luxfg (commented):
looks good to me, safe for the syntactical differences with the new version.

- f32 vs f<32> (still being discussed)
- domain<?,?,?> vs cartesian<?,?,?> (still being discussed)
- interval as array [interval, interval, interval] or more like a dictionary [x: interval, y:interval, z:interval] (not sure what I like better)
- spst prefixes to all types and operations

We can resolve those later also.

PR #3, opened 2024-02-20
ADD: Materialize operation, Examples

luxfg:
I added the materialize operation, which implements the idea from yesterday with the type cast.
I think calling it materialize is correct, because that is what it corresponds to.
By default, intermediate fields are never materialized, and always we recompute.
The materialize instruction breaks this up.

This is better than a type cast because it allows us to derive the types of the materialize with the type inference.
So we can always first insert the materialize disregarding the types and then infer them.

I also made some changes to the syntax to make things more consistent. It is a bit verbose, but I think more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ spst.field<spst.cartesian<?,?,?>, spst.extent<(?, ?, ?)>, f<32>> -> spst.field<s
 Performing type inference on the extents results in the following:
 
 ```mlir
-// Version 1: Inferred extents.
+// Version with inferred extents.
 %res = spst.computation (%in) 
 {
     schedule: spst.schedule<PARALLEL>,
@@ -407,7 +407,7 @@ This indicates a schedule where no recomputation is done.
 Instead, the values are explicitly communicated.
 
 ```
-// Version 2, with cast_extent to prevent type-propagation
+// Version with materialize to prevent recomputation
 %res = spst.computation (%in) 
 {
     schedule: spst.schedule<PARALLEL>,
--- a/README.md
+++ b/README.md
@@ -48,15 +48,15 @@ Note that not all targets might support all scalar types natively.
 For every triple of positive integers `x, y, z`, there is a domain type
 
 ```
-stencil.domain<x, y, z>
+spst.cartesian<x, y, z>
 ```
 
 If one or more dimensions of the domain are unknown,
 they may be replaced with a `?` placeholder:
 ```
-stencil.domain<?, ?, ?>
-stencil.domain<x, ?, ?>
-stencil.domain<x, y, ?>
+spst.cartesian<?, ?, ?>
+spst.cartesian<x, ?, ?>
+spst.cartesian<x, y, ?>
 ...
 ```
 The purpose of placeholders is to allow type-inference to deduce the
@@ -65,10 +65,12 @@ of the iteration domain have been inferred at compile time.
 
 ### Extent types
 
+TODO: Opening paragraph.
+
 An extent defines the access offsets of a stencil operation.
 For every integer triple `i, j, k` there is an extent access:
 ```
-stencil.access<i, j, k>
+stencil.access<i, j, k> ::= (i, j, k)
 ```
 If an extent is unknown, it can contain the placeholder extent `?`. The purpose of placeholders is to allow
 type-inference to deduce the extent accesses. We require the extent accesses to be compile-time inferrable.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,15 +48,15 @@ Note that not all targets might support all scalar types natively.
...
+spst.cartesian<x, y, z>
 ```
 
 If one or more dimensions of the domain are unknown,
 they may be replaced with a `?` placeholder:
 ```
-stencil.domain<?, ?, ?>
-stencil.domain<x, ?, ?>
-stencil.domain<x, y, ?>
+spst.cartesian<?, ?, ?>
```
tbennun, line 62: I'm not sure if we want `domain<cartesian, ?, ?, ?>`  (MLIR-style) or simply `cartesian<?, ?, ?>`. If you want the latter, we should probably define a base type for domain.

README.md
```diff
@@ -196,70 +215,229 @@ That is, any extent access plus an interval must not go beyond the bounds of the
...
+point of the domain, as indicated by the mask.
 
 Example:
 
 ```
-%x = stencil.if (%mask): field [D, extent[(0,0,0)], bool] -> (field [D, E_1, T_1]) {
+%x = spst.if (%mask): spst.field<D, extent<(0,0,0)>, bool> -> spst.field<D, E_1, T_1> {
   %x_1 = ...
-  stencil.yield %x_1 : field [D, E_1, T_1]
+  spst.yield (%x_1) : spst.field<D, E_1, T_1>
```
tbennun: Not sure yield is necessary if you're living outside of MLIR. It's only defined this way  in MLIR because `return` would have multiple meanings.

README.md
```diff
@@ -196,70 +215,229 @@ That is, any extent access plus an interval must not go beyond the bounds of the
...
+A direct translation results in the following computation:
+
+```mlir
+// Version with ? extents.
+%res = spst.computation (%in) 
+{
+    schedule: spst.schedule<PARALLEL>,
+    interval: [x: spsp.interval<?, ?> , y: spsp.interval<?, ?>, z: spst.interval<0, None>]
+} :
+spst.field<spst.cartesian<?,?,?>, spst.extent<(?, ?, ?)>, f<32>> -> spst.field<spst.cartesian<?,?,?>, spst.extent<(?, ?, ?)>, f<32>> {
```
tbennun: I think the tags in `f<32>` are redundant. Especially since there are other types such as `f16`, `bf16`, `tf32` and others that won't follow this representation.

tbennun (commented):
Name and operation definition of materialize LGTM. Left comments about the syntax in general

README.md
luxfg: yeah I think id' go for the latter and make a base type

README.md
luxfg: I wanted it to be as consistent as possible, so in f<i> the i is a parameter that specifies the bit-width. If a type like bool does not have a bit width parameter it would just be bool<> or bool for short.
I want to have it as a parameter so that ideally in the IR, I would like to have the same class for all floats, just with a different width parameter. No need to have so many classes for the same thing. Also all float<i> have the same semantics, so having the same class / type of IR node makes sense.

README.md
luxfg: I need to think about it a bit more. But I think ideally we would use the same thing within expressions and here.
so either both yield or both return.

README.md
tbennun: There are two reasons for why IRs generally prefer using something like `float`, `single`, or `f32` over your suggestion: 
1. If the syntax it describes a floating point type, f<32> would be incomplete.
2. `f32` also means the IEEE 754 floating point with 32-bit (single) precision. This also encapsulates behavior such as NaNs and denormalized numbers, as well as special values such as infinity.

If you really wish to have a canonical floating point with a certain bit-width (ignoring anything else other than IEEE standards, e.g., where you can infer the exponent range), what you need is `f<8,23>` for exponent and mantissa (and maybe `uf<e,m>` for unsigned floats, if they even exist). This would cover fp16 (`f<5,10>`), bfloat16 (`f<8,7>`), and e4m3/e5m2 fp8 variants, among others. It can seem misleading to write `f<8,23>` and thus one could alias it or define a shorthand to `f32`.

luxfg: ok. todo double check with mlir.

luxfg: todo: switch to f32

README.md
luxfg: I simplified everything to spst.return. I think it should be fine if we force all computations to only return in the end (not from within an if block). Not sure if this is consistent with gt4py though.

luxfg:
I addressed the comments and made an example, can you check if it's ok now for you?

PR #4, opened 2024-03-08
Model Placements and Costs

luxfg:
Prototypes the stencil program graph model and placement costs.

stencil_graph

```
Interleaved placement
Placement(offsets=array([[0, 0],
       [0, 0],
       [0, 0],
       [0, 1],
       [0, 1],
       [0, 1],
       [0, 0],
       [0, 0],
       [0, 0]], dtype=int32), strides=array([[2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1]], dtype=int32))
distances [2. 2. 2. 2. 0. 1. 2. 0.]
max distances 5.0
energy: 629211136
contention: 963.0

Separated placement
Placement(offsets=array([[  0,   0],
       [  0,   0],
       [  0,   0],
       [256,   0],
       [256,   0],
       [256,   0],
       [  0,   0],
       [  0,   0],
       [  0,   0]], dtype=int32), strides=array([[1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1]], dtype=int32))
distances [  1.   1.   1.   1.   0. 256.   1.   0.]
max distances 258.0
energy: 436207616
contention: 963.0

```


spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,133 @@
...
+        assert self.shape.shape[1] == 3
+        assert self.shape.shape[0] >= 0
+        assert np.issubdtype(self.shape.dtype, np.integer)
+
+    def __eq__(self, other):
+        return np.array_equal(self.shape, other.shape)
+
+    def is_horizontal(self) -> bool:
+        """
+        A stencil is horiontal if all z values (3rd coordinate) are 0
```
tbennun:
```suggestion
        A stencil is horizontal if all z values (3rd coordinate) are 0
```

scripts/placement_demo.py
```diff
@@ -0,0 +1,112 @@
...
+
+def demo_placement_interleave():
+    # Create a placement with two partitions, one with u, v, w, a, b, c and the other with x, y, z
+    # The partitions are placed on the same processing elements
+    # as numbers, we get
+    # u = 0, v = 1, w = 2, x = 3, y = 4, z = 5, a = 6, b= 7, c = 8
+    # we use offset (0, 0) for the first partition
+    # and offset (0, 1) for the second partition
+    # the stride is (2, 1) for both partitions
+    offsets = np.array([[0, 0], [0, 0], [0, 0], [0, 1], [0, 1], [0, 1], [0, 0], [0, 0], [0, 0]], dtype=np.int32)
```
tbennun: so, to clarify, a split node would exist as two separate fields with different domains? Because if so, currently FieldDomain does not allow such a breakdown.

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,133 @@
...
+from enum import Enum
+from typing import Sequence
+import igraph as ig
+import numpy as np
+
+
+@dataclass
+class FieldDomain:
+    """
+    Represents the 1x3 domain of a field
```
tbennun: why is this array 2-dimensional rather than 1? the domain at shape[0] will always be 1 regardless of the problem's dimensionality.

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,133 @@
...
+@dataclass
+class StencilShape:
+    # Represented as a k by 3 array
+    shape: np.ndarray
+
+    # Check if a stencil is valid
+    # A stencil is valid if it is a k times 3 shaped array and all its elements are integers
+    def __post_init__(self):
+        assert self.shape.shape[1] == 3
+        assert self.shape.shape[0] >= 0
```
tbennun: likely an unnecessary assertion

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,133 @@
...
+            if edge['direction'] == StencilDirection.BACKWARD:
+                assert np.all(edge['stencil'].shape[:, 0:2] == 0)
+                assert np.all(edge['stencil'].shape[:, 2] > 0)
+
+    def edges(self):
+        return self.graph.es
+
+    def plot(self):
+        # Plot the stencil graph
+        layout = self.graph.layout_reingold_tilford()
```
tbennun: shouldn't this algorithm only work for trees rather than general DAGs?

spatialstencil/placement/model.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def input_contention_of_fields(self, fields):
+        """
+        The input contention of a field is the sum of the communication volumes of its incoming edges.
+        :param fields: list of fields
+        :return: float
+        """
+        contention = 0.0
+        for v in fields:
+            for e in self.stencil_graph.graph.es.select(_target=v):
```
tbennun: annoying that there is no `in_edges` API.

spatialstencil/placement/model.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def output_contention_of_fields(self, fields):
+        """
+        The output contention of a field is the sum of the communication volume of its outgoing edges.
+        :param fields: list of fields
+        :return: float
+        """
+        contention = 0.0
+        for v in fields:
+            for e in self.stencil_graph.graph.es.select(_source=v):
```
tbennun: same for `out_edges`

tbennun (commented):
Some intermediate comments

spatialstencil/placement/graph.py
luxfg: While designed for trees, also works for DAGs. I changed it slightly to give better results on DAGs.

stencil_graph 2

scripts/placement_demo.py
luxfg: good point, I adapted the domain to be 2x3 with lower and upper bound

luxfg:
```
Placement(offsets=array([[0, 0],
       [0, 0],
       [0, 0],
       [1, 0],
       [1, 0],
       [1, 0],
       [1, 0],
       [1, 0],
       [0, 0]], dtype=int32), strides=array([[2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1],
       [2, 1]], dtype=int32))
Cost: PlacementCost(contention=769.0, energy_over_links=604.125, distance=6.0, depth=4.0, overall=789.0)

Separated placement
Placement(offsets=array([[  0,   0],
       [  0,   0],
       [  0,   0],
       [256,   0],
       [256,   0],
       [256,   0],
       [256,   0],
       [256,   0],
       [  0,   0]], dtype=int32), strides=array([[1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1],
       [1, 1]], dtype=int32))
Cost: PlacementCost(contention=769.0, energy_over_links=436.0, distance=514.0, depth=4.0, overall=970.0)

```

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,69 @@
...
+        assert len(stencils) == len(graph.es)
+        assert len(field_domains) == len(graph.vs)
+        self.graph.vs[StencilGraph.DOMAIN] = field_domains
+        self.graph.es[StencilGraph.STENCIL] = stencils
+        self.graph[StencilGraph.DOMAIN] = domain
+        # Check that for all forward edges the xy directions are 0 and the z direction is negative
+        for i, edge in enumerate(self.graph.es):
+            if edge[StencilGraph.STENCIL].direction == StencilDirection.FORWARD:
+                assert np.all(edge[StencilGraph.STENCIL].shape[:, 0:2] == 0)
+                assert np.all(edge[StencilGraph.STENCIL].shape[:, 2] < 0)
```
tbennun: This is not necessarily true. Read-only fields may have any offset. The written field may only adhere to this constraint though. Vertical advection demonstrates this.

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,69 @@
...
+        # Check for all backward edges the xy directions are 0 and the z direction is positive
+        for i, edge in enumerate(self.graph.es):
+            if edge[StencilGraph.STENCIL].direction == StencilDirection.BACKWARD:
+                assert np.all(edge[StencilGraph.STENCIL].shape[:, 0:2] == 0)
+                assert np.all(edge[StencilGraph.STENCIL].shape[:, 2] > 0)
```
tbennun: same as above

luxfg:
h_diffusion


horizontal diffusion

scripts/examples.py
```diff
@@ -0,0 +1,83 @@
...
+            cond_y = res_y[0, 0, 0] * (in_field[0, 1, 0] - in_field[0, 0, 0])
+            fly_field = 0 if cond_y > 0 else res_y
+
+            out_field = in_field[0, 0, 0] - coeff[0, 0, 0] * (
+                flx_field[0, 0, 0] - flx_field[-1, 0, 0] + fly_field[0, 0, 0] - fly_field[0, -1, 0]
+            )
+    """
+
+    # Define the domain
+    domain = FieldDomain(np.array([[0, 0, 0], [256, 256, 50]], dtype=np.int32))
```
tbennun: I'm used to 80 vertical points here rather than 50

scripts/examples.py
```diff
@@ -0,0 +1,83 @@
...
+        Stencil(np.array([[0, 0, 0]], dtype=np.int32), StencilDirection.PARALLEL),
+        # flx_field -> out_field
+        Stencil(np.array([[0, 0, 0], [-1, 0, 0]], dtype=np.int32), StencilDirection.PARALLEL),
+        # fly_field -> out_field
+        Stencil(np.array([[0, 0, 0], [0, -1, 0]], dtype=np.int32), StencilDirection.PARALLEL),
+    ]
+    # 11 vertices in_field = 0, out_field = 1, coeff = 2, lap_field = 3, res_x = 4, cond_x = 5, flx_field = 6,
+    # res_y = 7, cond_y = 8, fly_field = 9, out_field = 10
+    g = ig.Graph(directed=True, n=11, edges=[
+        (0, 3), (3, 4), (4, 5), (0, 5), (4, 6), (5, 6), (3, 7), (7, 8), (0, 8), (7, 9), (8, 9), (0, 1), (2, 1), (6, 10), (9, 10)
```
tbennun: why does out_field appear as both field 1 and 10 in the last 4 edges?

spatialstencil/placement/model.py
```diff
@@ -0,0 +1,301 @@
...
+        :return: float
+        """
+        # Calculate number of communication links (bidirectional)
+        domain = self.stencil_graph.domain()
+        number_of_partitions = placement.unique_offsets().shape[0]
+        grid_size = domain.xy_plane_area() * number_of_partitions
+        # TODO Note that for n_x by 1 or 1 by n_y domains, the number of links is smaller
+        # Note that for the energy computation, effects around the border of the domain are neglected
+        # This is an approximation that is valid for large domains
+        number_of_links = grid_size * 8
```
tbennun: would benefit some more "in-code" documentation - i.e., use 2 * 4 instead of 8

spatialstencil/placement/graph.py
luxfg: very good point! I missed this. This also means we can have "3d" stencils in addition to horizontal/vertical stencils.

luxfg:
v_advection

One thing I noticed in the vertical advection example is that we have something that acts like a cycle, where the datacol gets set by the data_col one plane above, which equals datacol. In the picture above I removed the cycle by making a new version from data_col, but data_col_2 should actually alias data_col I think. Another way to model it would be with a cycle, which messes up the cost computations.

In this example, it's easy to see that there is no communication in this cycle so we could contract it, but in general I am not sure what is possible and how to deal with such cycles. I need to think about it more.

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,199 @@
...
+    DOMAIN = 'domain'
+    STENCIL = 'stencil'
+    FIELD_NAME = 'name'
+    FIELD_VERSION = 'version'
+
+    def __init__(self, graph: ig.Graph,
+                 domain: FieldDomain,
+                 field_domains: Sequence[FieldDomain],
+                 field_names: Sequence[str],
+                 field_versions: Sequence[int],
```
tbennun: you can make that default to None and in that case fill it with zeroes.
```suggestion
                 field_versions: Optional[Sequence[int]] = None,
```

spatialstencil/placement/graph.py
```diff
@@ -0,0 +1,199 @@
...
+                 stencils: Sequence[Stencil]
+                 ) -> None:
+
+        self.graph = graph
+        assert len(stencils) == len(graph.es)
+        assert len(field_domains) == len(graph.vs)
+        assert len(field_names) == len(graph.vs)
+        assert len(field_versions) == len(graph.vs)
+        self.graph.vs[StencilGraph.FIELD_NAME] = field_names
+        self.graph.vs[StencilGraph.FIELD_VERSION] = field_versions
```
tbennun:
```suggestion
        self.graph.vs[StencilGraph.FIELD_VERSION] = field_versions or ([0] * len(field_names))
```

PR #5, opened 2024-06-13
AST -> Stencil IR

tbennun:
Implements a parser for GT4Py embedded Python DSL, a parser/printer for the Stencil IR, and lowering from GT4Py to the Stencil IR

Components:
- [x] GT4Py syntax parser
   - [x] Conditionals
- [x] Stencil IR object-oriented definition
  - [x] Lark support
- [x] IR node traversal and transformation (GT4Py, Stencil IR)
- [x] Lowering GT4Py AST to Stencil IR
  - [x] Field versioning
  - [x] Constant propagation
  - [x] Type inference
     - [x] Input/output
     - [x] Scalar type
     - [x] Extents
     - [x] Domain
  - [x] Materialize everything (as IR node transformation)

Stencils:
- [x] hdiff
- [x] vadv
- [x] hdiffsa
- [x] hdiffsmag
- [x] uvbke


tbennun:
Ready for review @glukas

irspec/docs/stencil/stencil.md
```diff
@@ -37,6 +37,15 @@ bool
...
 Note that not all targets might support all scalar types natively.
 
+Type promotion rules (e.g., when adding an `f16` to `f32`) follows the ANSI C standard, and evaluates
+the following rules in this order:
+
+  * If one type is integral and another is floating point, the result will be floating point;
+    * Example: `i32 + f16 -> f16`
+  * When one of the types is wider, the widest type will be used for the output;
+    * Example: `i16 + u16 -> u16`
+  * For the purposes of type promotion, `bool` is considered a signed integral 1-bit value.
```
luxfg: is bool signed 1 bit or unsigned 1 bit?

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+
+    def as_ir(self, indent: int = 0) -> str:
+        """
+        Returns the node as a parseable Stencil IR version.
+
+        :param indent: Indentation for the IR node.
+        """
+        return str(self)
+
+    def validate(self) -> None:
```
luxfg, line 79: Arguably validate should be part of BaseNode

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+        """
+        Runs assertions on the node.
+        """
+        pass
+
+    def pretty(self) -> str:
```
luxfg: Also pretty could probably be part of BaseNode

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+        """
+        Pretty-prints the contents of this AST node.
+        """
+        return pprint.pformat(self)
+
+
+class Domain(Node):
```
luxfg: Could it make sense to have a class or interface 'BaseType' and have all types inherit from that? 
This could for example contain the 'is_subtype_of' / 'is_supertype_of' this could allow some generic type checking.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,160 @@
...
+                for item in field:
+                    if isinstance(item, BaseNode):
+                        yield item
+
+
+# Create a generic type T that extends the base node type
+BaseNodeT = TypeVar('BaseNodeT', bound=BaseNode)
+
+
+def walk(node: BaseNode):
```
luxfg: could be a method of BaseNode?

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,160 @@
+"""
+Provides basic AST/IR utilities for traversal, transformation, and find/replacement.
```
luxfg: helpers is an undescriptive name. Consider changing to irnode.py.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,160 @@
...
+Walking and transformation are based on Python's ``ast`` module, but introduce additional
+functionality such as IR language testing and dataclass support.
+"""
+import ast
+from dataclasses import dataclass
+from typing import Generic, TypeVar
+
+
+@dataclass
+class BaseNode:
```
luxfg: I think we should have 'Operation', 'Type', and 'BasicBlock' as interfaces or abstract base classes defined maybe in this file so we can reuse them across different IRs.

luxfg: The BasicBlock could just contain a list of operations.
We will also need the successors/predecessors of basic blocks later for the spatial ir, but we could do that in a subclass of basic block.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+    def generic_visit(self, node: BaseNodeT):
+        """
+        Recursively visit all children of the node.
+        Called if no explicit visitor function exists for a node.
+        If a visitor function exists for a given
+        node type, it MUST call ``self.generic_visit(node)`` to visit its children.
+        """
+        if isinstance(node, BaseNode):
+            for _, value in node.iter_fields():
+                if isinstance(value, list):
```
luxfg: I am thinking if it's worth it to have a generic class NodeSequence[T] that provides a slightly more safe and extensible way to represent lists of nodes. It could have a builtin validate that ensures that all its elements conform to T. I could imagine that if we need to add more functionality on handling lists of nodes, this could be helpful.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+
+class IRNodeTransformer(IRNodeVisitor[BaseNodeT]):
+    """
+    A ``NodeVisitor`` subclass that walks the AST/IR and allows modification of nodes.
+    See ``ast.NodeTransformer`` for more information.
+    """
+
+    def generic_visit(self, node: BaseNodeT):
+        for field, old_value in node.iter_fields():
+            if isinstance(old_value, list):
```
luxfg: what about tuples?

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+                new_values = []
+                for value in old_value:
+                    if isinstance(value, BaseNode):
+                        value = self.visit(value)
+                        if value is None:
+                            continue
+                        elif not isinstance(value, BaseNode):
```
luxfg: So this assumes that it returns a list or tuple or other sequence if it does not return a BaseNode?
I would at least add some assertion to that regard to clarify the intent. looks brittle.
Maybe the above idea for using a predefined NodeSequence interface/class could help, so we could actually match against that instead of doing a deduction by exclusion.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+    Finds and replaces a name with another value on Python ASTs.
+    """
+
+    def __init__(self, repldict: dict[str, ast.AST]):
+        """
+        Creates a find-and-replace AST node transformer.
+
+        :param repldict: A dictionary mapping a source name to a target replacement AST node.
+        """
+        self.replace_count = 0
```
luxfg: what is the use of replace_count? It is not really exposed as part of the interface of the class. either declare "nicely" + document or remove.

spatialstencil/syntax/stencil_ir/language.lark
```diff
@@ -25,30 +25,35 @@ unknown_dim_literal : "?"
...
+?enum_literal : schedule_type
 
 // Language built-in types
-?dim : integer_literal | unknown_dim_literal
+?dim : posneg_integer_literal | unknown_dim_literal
+!dim_or_end : dim | "None"
 dim_list: dim ("," dim)*
-extent_tuple : "(" dim ("," dim)* ")"
+extent_offset_tuple : "(" dim ("," dim)* ")"
+extent_interval_tuple : "[" dim ":" dim ("," dim ":" dim)* "]"
```
luxfg: I would suggest to unify this with interval type.

luxfg: also shouldnt it allow dim_or_end?

spatialstencil/syntax/stencil_ir/language.lark
```diff
@@ -25,30 +25,35 @@ unknown_dim_literal : "?"
...
+extent_tuple : extent_offset_tuple ("in" extent_interval_tuple)?
 extent_tuple_list : extent_tuple ("," extent_tuple)*
-extent_type : "extent" "<" extent_tuple_list ">"
-domain_type : "domain" "<" dim_list ">"
-field_type : "field" "<" domain_type "," extent_type "," standard_type ">"
-interval_type : "interval" "<" (dim | "None") "," (dim | "None") ">"
+extent_type : ("spst.")? "extent" "<" extent_tuple_list ">"
+domain_type : ("spst.")? ("domain" | "cartesian") "<" dim_list ">"
+field_type : ("spst.")? "field" "<" domain_type "," extent_type "," standard_type ">"
+interval_type : ("spst.")? "interval" "<" dim_or_end "," dim_or_end ">"
```
luxfg:
```
interval_tuple_short: dim_or_end ":" dim_or_end
interval_type : ("spst.")? "interval" "<" dim_or_end "," dim_or_end ">" | interval_tuple_short
```

spatialstencil/syntax/stencil_ir/language.lark
```diff
@@ -25,30 +25,35 @@ unknown_dim_literal : "?"
...
-extent_tuple : "(" dim ("," dim)* ")"
+extent_offset_tuple : "(" dim ("," dim)* ")"
+extent_interval_tuple : "[" dim ":" dim ("," dim ":" dim)* "]"
+extent_tuple : extent_offset_tuple ("in" extent_interval_tuple)?
 extent_tuple_list : extent_tuple ("," extent_tuple)*
-extent_type : "extent" "<" extent_tuple_list ">"
-domain_type : "domain" "<" dim_list ">"
-field_type : "field" "<" domain_type "," extent_type "," standard_type ">"
-interval_type : "interval" "<" (dim | "None") "," (dim | "None") ">"
+extent_type : ("spst.")? "extent" "<" extent_tuple_list ">"
```
luxfg:
```
extent_type_short : "{" extent_tuple_list "}"
extent_type: ("spst.")? "extent" "<" extent_tuple_list ">" | extent_type_short
```

spatialstencil/syntax/stencil_ir/language.lark
```diff
@@ -25,30 +25,35 @@ unknown_dim_literal : "?"
...
+domain_type : ("spst.")? ("domain" | "cartesian") "<" dim_list ">"
```
luxfg: domain_type_short : "["dim_list "]"
domain_type: ("spst.")? ("domain" | "cartesian") "<" dim_list ">" | domain_type_short

irspec/docs/stencil/stencil.md
```diff
@@ -62,48 +74,95 @@ The purpose of placeholders is to allow type-inference to deduce the
 domain sizes. To lower the representation, we require all dimensions
 of the iteration domain have been inferred at compile time.
 
+!!! abstract "Domain Shorthand Notation"
+    When clear from context, the shorthand notation `[x, y, z]` may be used instead.
+
+    For example, `spst.cartesian<1, ?, 3>` is equivalent to `spst.cartesian<1, ?, 3>`
```
tbennun: What does this mean?

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+
+    def as_ir(self, indent: int = 0) -> str:
+        return f'spst.field<{self.domain.as_ir()}, {self.extent.as_ir()}, {self.dtype.as_ir()}>'
+
+
+DataType = FieldType | ScalarType
+
+
+@dataclass
+class TypeInfo(Node):
```
luxfg: Maybe better named as 'OperationType'

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+        return f'({self.value.as_ir(indent)})'
+
+
+class Operation:
+    """
+    Interface for an operation.
+
+    An Operation *must* have a field called ``typeinfo`` defined.
+    """
+    typeinfo: TypeInfo
```
luxfg: Maybe rename typeinfo to 'operation_type'

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+@dataclass
+class StatementBlock(Node, Operation):
+    """
+    A single statement in a Stencil IR computation.
+    """
+    outputs: list[Identifier]
+    inputs: list[Identifier]
+    attributes: list[tuple[str, Node]]
```
luxfg: todo: double check how this behaves with the visitors, what is the desired behavior when visiting attributes?

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+@dataclass
+class IfBlock(Node, Operation):
+    """
+    If/elif/else block operating on a mask tensor.
+    """
+    outputs: list[Identifier]
+    condition: Identifier
+    typeinfo: TypeInfo
+    body: list[StatementBlock | ReturnOp]
+    else_ifs: list[tuple[Identifier, list[StatementBlock | ReturnOp]]] | None  # List of (condition, body)
```
luxfg: here, we definitely want to recurse into the elements of the else_ifs in the visitors, and they are tuples.

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -0,0 +1,541 @@
...
+
+@dataclass
+class ComputationBlock(Node, Operation):
+    """
+    A computational block with an interval in a stencil IR computation.
+    """
+    outputs: list[Identifier]
+    inputs: list[Identifier]
+    schedule: ComputationType
+    interval: tuple[Interval, Interval, Interval]
```
luxfg: here, I also think we would want to recurse into the interval when visiting, so tuples should be handled accordingly.

luxfg (changes requested):
I have reviewed helpers.py, irnodes.py, and the docs in detail. I will continue later / tomorrow with the rest.

irspec/docs/stencil/stencil.md
luxfg: I fixed it.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: agreed, should work now.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: should work now, thanks!

spatialstencil/syntax/stencil_ir/language.lark
tbennun: > also shouldnt it allow dim_or_end?

Fixed.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: should work now, though we could add `Attribute` as an explicit class if you wish to visit a specific attribute and access its name, e.g., when writing a node visitor.

spatialstencil/syntax/helpers.py
tbennun: Fixed.

spatialstencil/syntax/helpers.py
tbennun: Fixed.

spatialstencil/syntax/helpers.py
tbennun: Fixed

spatialstencil/syntax/helpers.py
tbennun: It is used to make sure something was replaced in some passes (field versioning I think?). I can document but cannot remove without adding another O(n) pass on the statements with yet a second visitor.

spatialstencil/syntax/helpers.py
tbennun: I would call it Block rather than basic block (which would be used for a lower-level LLVM IR BB).

spatialstencil/syntax/helpers.py
tbennun: Maybe node.py? Some nodes are AST nodes.

spatialstencil/syntax/helpers.py
tbennun: I copied the interface in Python's `ast` module. We can make it a method.

spatialstencil/syntax/stencil_ir/type_inference.py
```diff
@@ -0,0 +1,559 @@
...
+class ExtentAssigner(sast.NodeTransformer):
+    """
+    Sets extents based on given dictionary.
+    """
+
+    def __init__(self, field_extents: dict[str, dict[tuple[int | None], set[tuple[int | None]]]]):
+        super().__init__()
+        self.field_extents = field_extents
+
+    def _sort_extents(self, extent_set: dict[tuple[int | None], set[tuple[int | None]]]):
```
luxfg: I am generally wondering about canonicalization. Should nodes have a method to canonicalize themselves? or should they be able to check if they are in canonical form?

spatialstencil/syntax/stencil_ir/type_inference.py
```diff
@@ -0,0 +1,559 @@
...
+        names.add(k.name)
+        if k.name not in versions:
+            versions[k.name] = k.version
+        else:
+            versions[k.name] = func(k.version, versions[k.name])
+
+    return [sast.Identifier(name, versions[name]) for name in sorted(names)]
+
+
+class ExtentAssigner(sast.NodeTransformer):
```
luxfg, line 471: (Ok, I see now that we get some boilerplate code for ensuring the types are updated everywhere, I think it's a reasonable solution using the Assigners.)

spatialstencil/syntax/stencil_ir/analysis.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self._DEFAULT_INTERVAL = (sast.Interval(0, None), sast.Interval(0, None), sast.Interval(0, None))
+
+        self.extents: dict[str, dict[tuple[int | None], set[tuple[int | None]]]] = defaultdict(lambda: defaultdict(set))
+        self.interval: tuple[sast.Interval, sast.Interval, sast.Interval] = self._DEFAULT_INTERVAL
+
+    @property
+    def flat_interval(self) -> tuple[int | None]:
```
luxfg: Expected type 'tuple[int | None]', got 'tuple[int | Literal["?"] | None, ...]' instead

spatialstencil/syntax/stencil_ir/analysis.py
```diff
@@ -0,0 +1,122 @@
...
+class ExtentCollector(sast.NodeVisitor):
+    """
+    A node visitor that collects all input and output extents from field accesses in the visited blocks/statements.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._DEFAULT_INTERVAL = (sast.Interval(0, None), sast.Interval(0, None), sast.Interval(0, None))
+
+        self.extents: dict[str, dict[tuple[int | None], set[tuple[int | None]]]] = defaultdict(lambda: defaultdict(set))
```
luxfg, line 18: the types of self.extents are not correct here, I think the tuples should be different. A 3 tuple has type tuple[int | None, int | None, int | None]

spatialstencil/syntax/stencil_ir/analysis.py
```diff
@@ -0,0 +1,122 @@
+"""
+Analysis passes on the Stencil IR.
+"""
+from collections import defaultdict
+from spatialstencil.syntax.stencil_ir import irnodes as sast
+
+
+class ExtentCollector(sast.NodeVisitor):
```
luxfg, line 9: The type checking is not consistent with the specification:

> The expression may only contain accesses to the fields `in_1, ... in _n` at the allowed extents.
> For any field, only accesses that are within their extent type are permitted within the statements.
> Which accesses are permitted is relative to the output's extent type.
> Specifically, every access to `(i, j, k)` is added to every output extent `(x, y, z)`
> and results in an access to `(i+x, j+y, k+z)`.
> 
> This corresponds to a Minkovski sum of the accesses of the statements and the accesses of the output's extent type._
> The arguments to the statement must be a sub-type of the union of all these accesses.
> Specifically, accessing a field with placeholders in their extent type always type checks.
> 


Specifically, if the output extent uses a non-zero offset access of %x, then this means we need to compute %x for additional extents.

spatialstencil/syntax/stencil_ir/analysis.py
```diff
@@ -0,0 +1,122 @@
...
+        self.outputs: set[sast.Identifier] = set()
+
+    # Visitors that return used identifiers
+    def visit_Identifier(self, node: sast.Identifier) -> set[sast.Identifier]:
+        return {node}
+
+    def visit_Expression(self, node: sast.Expression) -> set[sast.Identifier]:
+        if isinstance(node.value, (int, float)):
+            return set()
+        return self.visit(node.value)
```
luxfg, line 85: assert the type that node.value must have

luxfg (changes requested):
I reviewed the type inference. There is an inconsistency in the understanding of extent type inference.

spatialstencil/syntax/stencil_ir/analysis.py
luxfg: I added some tests for the extent inference below to showcase the difference

spatialstencil/syntax/helpers.py
luxfg: ok, in that case add doc

spatialstencil/syntax/helpers.py
luxfg: Block is good.

luxfg: Alternatively, if we have a generic NodeList class, we could just use that and I would be fine with it. So for example NodeList[AssignOp | ReturnOp]

spatialstencil/syntax/stencil_ir/irnodes.py
luxfg: ok. right now, it visits the nested nodes, but not the attributes themselves. This is fine I guess. Alternatively, we could also mandate the use of special NodeList and NodeTuple classes for nested nodes to make it more explicit what is visited and what is not. Right now, every Node nested somewhere within lists and tuples is visited.

spatialstencil/syntax/gt4py/astnodes.py
```diff
@@ -0,0 +1,132 @@
...
+    FieldI = enum.auto()
+    FieldJ = enum.auto()
+    FieldK = enum.auto()
+    int = enum.auto()
+    float = enum.auto()
+
+
+class GTree(helpers.BaseNode):
+
+    def pretty(self, indent: int = 0) -> str:
```
luxfg, line 33: printing should be in base node.

spatialstencil/syntax/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+            elif isinstance(value, (list, tuple)):
+                value = self.generic_visit_sequence(value)
+            new_values.append(value)
+
+        if isinstance(sequence, tuple):
+            return tuple(new_values)
+        return new_values
+
+
+class ASTFindReplace(ast.NodeTransformer):
```
luxfg: I think ASTFindReplace should be in its own file, as it does not belong to the BaseNode and it could be confused to think it is a IRNodeTransformer when it is not.

spatialstencil/syntax/helpers.py
luxfg: Ok, but the classes are called IRNodeVisitor, IRNodeTransformer. We can also split it up further, having a basenode.py and one or more *visitor(s).py

spatialstencil/syntax/stencil_ir/type_inference.py
luxfg: you can also see this issue in two of the new tests (test_infer_extent_no_materialize / test_infer_extent_materialize). I think we would want to be able to get a canonical representation so that we can ensure that the as_ir strings always match for programs that only differ in the order of sets.

spatialstencil/syntax/stencil_ir/analysis.py
luxfg: My rationale for this choice of definition is that we want to know which offsets are actually accessed, not just the values that appear in the syntax, which should make translation easier. I think it only shows up with no materialize for statements. In my view we do want to propagate the minkovski sum to the program arguments, so that means the program argument types are not necessarily equal to the types used by the statements, but are subtypes (see the example with materialize).

luxfg: Note that we will also need a pass that removes the non-materialized fields, so this could be a way around having to implement the extent inference for such fields (as we generally remove them before lowering). This pass will need a similar logic though as the type inference.

irspec/docs/stencil/stencil.md
tbennun: Fixed.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: yes, done. and you can use the built-in Python `issubclass` or `isinstance`.

spatialstencil/syntax/helpers.py
tbennun: Type is a reserved name in Python so I used `IRType`. We discussed NodeList so it stays list. Otherwise done.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: Fixed.

spatialstencil/syntax/stencil_ir/irnodes.py
tbennun: Fixed.

spatialstencil/syntax/stencil_ir/analysis.py
tbennun: If you specify a tuple with one type, it is a variable-length tuple of that type (as far as I'm aware)

spatialstencil/syntax/stencil_ir/analysis.py
tbennun: The recursive visit ensures we get the right type of result. I don't understand what needs to be done here.

spatialstencil/syntax/stencil_ir/type_inference.py
tbennun: Canonicalization is implemented, the tests still fail (probably due to local extents as discussed in another comment). However, this note is resolved.

luxfg (approved):
We can merge the PR. Note that the type inference is still not as intended when no materialize is present and the extent types of computations are also not taking into account the dependencies between statements. I think we can fix it later though as well and I want to figure out what we actually need for lowering exactly.

PR #10, opened 2024-08-23
Rewrite Extent & Domain Inference

luxfg:
I rewrote the extent & domain inference so that extents & domain accesses are "local" in scope.

That is, a type annotation on a statement means that the statement accesses a certain sub-domain of the field with offsets as described. To see the size of a field, one should look at the domain of the computation arguments or program arguments. These in turn do not carry meaningful offset information, hence we leave it un-inferred.
Because types are "local", we do not need the "in interval" notation for types anymore: Each computation has its own types that are bound to its scope.

The main workhorse of the analysis is a "Def-Use" analysis. Then, we work backwards through the program using a reversed-visitor, to which I add scoping information using a stack of scopes. This allows us to make the inferences using the general paradigm:

- Initialize outputs based on previous knowledge, if there are no uses.
- Consider all the in-scope uses of an output, and combine the information to deduce the result type.
- Recurse into the constuct, if necessary
- Construct input type based on local information and the result type

Note that the order of these steps and the specifics vary from op to op and analysis to analysis.

To make this work, we require proper return statements at the end of every computation and program. I have added them in most of the examples, but not yet updated the lowering from gt4py (probably faster if you do it).

With this new interpretation, we should be well equipped for lowering to spatial (or any other ir) because we have all the info to construct compute ranges (from statement types) and place ranges (from computation types & statement output types) and more. 

There are still some cornercases that I want to test more thoroughly, but most of the functionality works,  without materialize, with mutliple returns, and if-else constructs. The extent inference works with and without materialize, but I am still working on making domain inference work without materialize.

I also made some minor changes to the IR, mainly addressing some idiosyncrasies.


luxfg:
Note that we require all computations and programs to end with a return statement, so the gt4py lowering fails. @tbennun could you implement that change?

Adding returns makes all the analysis passes a lot nicer, because the returns make the use explicit.

samples/spst/if_else.spst
```diff
@@ -0,0 +1,32 @@
...
+    %qstar = spst.if (%pred1) : spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> -> spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> {
+      %qstar#1 = spst.statement (%f2) : spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> -> spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> {
+        spst.return %f2
+      }
+      spst.return %qstar#1 : field<[?,?,?], {(?, ?, ?)}, f32>
+    } elif (%pred2) {
+      %qstar#2 = spst.statement (%f2) : spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> -> spst.field<[?:?, ?:?, ?:?], {(?, ?, ?)}, f32> {
+        spst.return %f2[0, 1, 0]
+      }
+      spst.return %qstar#2 : field<[?,?,?], {(?, ?, ?)}, f32>
```
tbennun: There is a bit of an issue here with the version being incremented in a branch. Let's discuss.

samples/spst/laplacian_mat_ext_dom.spst
```diff
@@ -0,0 +1,35 @@
...
+{
+    %res = spst.computation (%in)
+    {
+        schedule = PARALLEL,
+        interval = [spst.interval<0, None> , spst.interval<0, None>, spst.interval<0, None>]
+    } : spst.field<spst.cartesian<-1:129,-1:130,0:80>, spst.extent<(?, ?, ?)>, f32>
+      -> spst.field<spst.cartesian<0:128,0:128,0:80>, spst.extent<(0, 0, 0)>, f32>
+    {
+        %out_1 = spst.statement(%in)
+                : spst.field<spst.cartesian<-1:129,-1:130,0:80>,
```
tbennun: There is a bit of a discrepancy with this one and how domain inference works (see `no_mat_ext_dom` version). When I read this it seems like it should only go to 130.

samples/spst/laplacian_no_mat_ext_dom.spst
```diff
@@ -0,0 +1,30 @@
...
+    %res = spst.computation (%in)
+    {
+        schedule = PARALLEL,
+        interval = [0:None, 0:None, 0:None]
+    } : field<[-1:129, -1:131, 0:80],
+       {(?, ?, ?)}, f32>
+      -> field<[0:128, 0:128, 0:80], {(0, 0, 0)}, f32>
+    {
+        %out_1 = spst.statement(%in)
+                : field<[-1:129,-1:131, 0:80], {(0, 0, 0), (-1, 0, 0), (1, 0, 0), (0, -1, 0), (0, 1, 0), (0, 1, 0), (-1, 1, 0), (1, 1, 0), (0, 2, 0)}, f32>
```
tbennun: Looks like this should be -1:130, but the test fails if this is not 131. I think this is incorrect and the code should be fixed so that it's -1:130

samples/spst/laplacian_no_mat_ext_dom.spst
```diff
@@ -0,0 +1,30 @@
+// Version with inferred extents
+// Without materialize
+%pres = spst.program (%in) {} :
+       field<[-1:129, -1:131, 0:80], {(?, ?, ?)}, f32>
```
tbennun: See below

spatialstencil/lowering/gt4py_to_stencil_ir.py
```diff
@@ -60,29 +62,31 @@ def field_versioning(program: gtast.GTProgram):
...
+
+                        old_name = stmt.target
+                        name_to_version[stmt.target] += 1
+                        stmt.target = f'{stmt.target}#{name_to_version[stmt.target]}'
+                        replacements[old_name] = ast.Name(id=stmt.target)
+                    else:
+                        names.add(stmt.target)
+                elif isinstance(stmt, gtast.GTIfStatement):
+                    # TODO Handle if statements
+                    pass
```
tbennun, line 99: I don't know whether we should increment the versions in the body of each if/elseif. Let's discuss. If we do increment the versions, this becomes a recursive traversal of the body and else(if) bodies.

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class DefUseAnalysis(sast.ScopedNodeVisitor):
+
+    """
+    For each field variable, constructs the uses of it,
+    storing the field type object and the enclosing scope of it in a dictionary.
+
+    Moreover, we construct for each use the defining identifier and the scope of the definition.
+    This assumes that the program is in SSA form, that is, each identifier is defined exactly once.
```
tbennun, line 26: So, in MLIR there is the notion of IsolatedFromAbove scopes. We may want to reuse these kinds of scopes here. Otherwise looks good.

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    def __init__(self,
+                 def_use: dict[sast.Identifier, list[ScopedUse]] = None,
+                 use_def: dict[sast.Identifier, ScopedDefinition] = None):
+        """
+        Initializes the def-use and use-def analysis.
+        :param def_use: If None, the def-use analysis is disabled.
```
tbennun, line 34:
```suggestion

        :param def_use: If None, the def-use analysis is disabled.
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+        self.def_use = def_use
+        self.use_def = use_def
+        self._current_scope = []
+
+    def add_use(self, node: sast.Identifier, field_type: sast.FieldType):
+        """
+        Adds a use of a field to the def_use dictionary, with the current scope.
+        :param node: The identifier
+        :param field_type: The field type object
+        :return:
```
tbennun:
```suggestion
        Adds a use of a field to the def_use dictionary, with the current scope.

        :param node: The identifier
        :param field_type: The field type object
```

tbennun: Notice how there must be an empty newline between the definition and param for Sphinx/readthedocs to work

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+
+        if node not in self.def_use:
+            self.def_use[node] = []
+        self.def_use[node].append(ScopedUse(self.get_scope(), field_type))
+
+    def add_definition(self, node: sast.Identifier, field_type: sast.FieldType):
+        """
+        Adds a definition of a field to the use_def dictionary, with the current scope.
+        :param node: The identifier
+        :return:
```
tbennun:
```suggestion
        Adds a definition of a field to the use_def dictionary, with the current scope.

        :param node: The identifier
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+        assert node not in self.use_def, f"Program is not in SSA form, {node} is defined multiple times"
+
+        self.use_def[node] = ScopedDefinition(self.get_scope(), field_type)
+
+    def visit_StatementBlock(self, node: sast.StatementBlock):
+        """
+        A statement uses all its argument types
+        and defines all its outputs.
+        :param node:
+        :return:
```
tbennun:
```suggestion
        A statement uses all its argument types and defines all its outputs.

        :param node: The node to visit.
```
Also this is more of a comment rather than a method documentation IMO.

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+        # WE INTENTIONALLY DO NOT RECURSE INTO THE STATEMENT BLOCK
+        # AS IT IS A LOCAL SCOPE
+
+    def visit_MaterializeOp(self, node: sast.MaterializeOp):
+        """
+        A materialize operation uses its argument
+        and defines its output.
+        :param node:
+        :return:
+        """
```
tbennun:
```suggestion
        """
        A materialize operation uses its argument and defines its output.

        :param node: The node to visit.
        """
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def visit_IfBlock(self, node: sast.IfBlock):
+        """
+        An if block uses its condition and the conditions of its if-else blocks.
+        The if-else blocks share the same type as the if block.
+
+        An if block defines its outputs.
+        :param node:
+        :return:
+        """
```
tbennun:
```suggestion
        """
        An if block uses its condition and the conditions of its if-else blocks.
        The if-else blocks share the same type as the if block.
        An if block defines its outputs.

        :param node:  The node to visit.
        """
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self.generic_visit(node)
+
+    def pre_visit_ComputationBlock(self, node: sast.ComputationBlock):
+        """
+        A computation block uses all its inputs.
+        and defines all its outputs.
+        :param node:
+        :return:
+        """
```
tbennun:
```suggestion
        """
        A computation block uses all its inputs and defines all its outputs.

        :param node: The node to visit.
        """
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+            self.add_use(arg_id, arg_t)
+        for out_id, out_t in zip(node.outputs, node.operation_type.destination):
+            self.add_definition(out_id, out_t)
+
+    def do_visit_ComputationBlock(self, node: ComputationBlock):
+        """
+        Within the computation block, all inputs are defined.
+        :param node:
+        :return:
+        """
```
tbennun:
```suggestion
        """
        Within the computation block, all inputs are defined.

        :param node: The node to visit.
        """
```

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+        for arg_id, arg_t in zip(node.inputs, node.operation_type.source):
+            self.add_definition(arg_id, arg_t)
+        self.generic_visit(node)
+
+    def visit_ReturnOp(self, node: sast.ReturnOp):
+        # A return uses all its arguments and does not define anything
```
tbennun, line 133: I would convert all prior docstrings to comments like this.

spatialstencil/syntax/stencil_ir/def_use_analysis.py
```diff
@@ -0,0 +1,152 @@
...
+        assert all(isinstance(arg.value, sast.Identifier) for arg in node.values)
+        # And create a use for each field access
+        for i in range(len(node.values)):
+
+            arg, arg_t = node.values[i].value, node.operation_type.source[i]
+
+            # TODO This is a hot-fix! Should be done elsewhere.
+            if isinstance(arg_t, sast.AnyType):
+                node.operation_type.source[i] = sast.FieldType.empty()
+                arg_t = node.operation_type.source[i]
```
tbennun: Why is this here if someone else fills the operation type before domain inference? Where does it fail if you remove this?

spatialstencil/syntax/stencil_ir/lark_to_ast.py
```diff
@@ -215,16 +241,14 @@ def program(self, args, meta=None):
...
 
 
 def _make_offset_and_interval(extent_tuple):
     if len(extent_tuple) == 2:
-        interval_list = []
-        for intvl in extent_tuple[1]:
-            interval_list.extend([intvl.start, intvl.end])
-        return irnodes.OffsetAndInterval(_make_dimtuple(extent_tuple[0]), tuple(interval_list))
-    return irnodes.OffsetAndInterval(_make_dimtuple(extent_tuple[0]))
+        assert False, "NOT SUPPORTED"
```
tbennun: Please `raise NotImplementedError` instead (standard Python)

spatialstencil/syntax/stencil_ir/lark_to_ast.py
```diff
@@ -215,16 +241,14 @@ def program(self, args, meta=None):
 
 # Helper functions
 def _make_dimtuple(tup):
-    return tuple((None if dim == '?' else dim) for dim in tup)
+    return tuple((dim for dim in tup))
```
tbennun:
```suggestion
    return tuple(tup)
```
Also this function is obviated now

spatialstencil/syntax/stencil_ir/lark_to_ast.py
```diff
@@ -140,10 +159,15 @@ def call(self, args, meta=None):
 
     # Operations
     def return_expr(self, args, meta=None):
-        if isinstance(args[-1], irnodes.OperationType):
-            return irnodes.ReturnOp(args[:-1], args[-1])
-        return irnodes.ReturnOp(args)
 
+        # Find the operation type arguments
+        optype = next((i for i in range(len(args)) if isinstance(args[i], irnodes.OperationType)), None)
```
tbennun, line 165: Why not check for the last argument? Shouldn't a args always take the type of [value1, ..., valueN, (OperationType)?]

tests/test_type_inference.py
```diff
@@ -30,20 +28,166 @@ def test_result_type_function(self):
...
+                program = parser.parse_file(f)
+
+            # extent inference
+            extent_inference.infer_field_extents(program)
+
+            # Assert all statements have extents
+            for computation in program.computations:
+                if isinstance(computation, ComputationBlock):
+                    for statement in computation.body:
+                        if isinstance(statement, StatementBlock) or isinstance(statement, MaterializeOp):
```
tbennun:
```suggestion
                        if isinstance(statement, (StatementBlock, MaterializeOp)):
```

tests/test_type_inference.py
```diff
@@ -30,20 +28,166 @@ def test_result_type_function(self):
...
+
+        # Check that all domains have been inferred
+        for computation in program.computations:
+            if isinstance(computation, ComputationBlock):
+                for statement in computation.body:
+                    for input in statement.operation_type.source:
+                        if isinstance(input, ScalarType):
+                            continue
+                        assert not input.domain.is_unknown()
+                    if isinstance(statement, StatementBlock) or isinstance(statement, MaterializeOp):
```
tbennun:
```suggestion
                    if isinstance(statement, (StatementBlock, MaterializeOp)):
```

tests/test_type_inference.py
```diff
@@ -30,20 +28,166 @@ def test_result_type_function(self):
...
+
+        assert result == golden_result
+
+
+    def assert_infer_domains(self, program: Program):
+        """
+        Asserts that all domains have been inferred
+        :param program:
+        :return:
+        """
```
tbennun:
```suggestion
        """
        Asserts that all domains have been inferred. Raises assertion exceptions on failure.

        :param program: The program to test.
        """
```

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+        for b in node.body:
+            self.visit(b)
+    def visit_Identifier(self, node: sast.Identifier):
+        # If a bare identifier (i.e., no subscript) is used, the extent (0, 0, 0) should be added
+        self.extents[node.name].add(sast.Offset((0, 0, 0)))
+
+    def visit_Subscript(self, node: sast.Subscript):
+        # If a subscript is found, add its subscript to the extents.
+        # Make sure not to recursively visit into the subscript to avoid adding (0, 0, 0)
+        self.extents[node.value.name].add(sast.Offset((node.subscript[0], node.subscript[1], node.subscript[2])))
```
tbennun:
```suggestion
        self.extents[node.value.name].add(sast.Offset((node.subscript[0], node.subscript[1], node.subscript[2])))

```

tbennun: Add newline to end of file

spatialstencil/syntax/stencil_ir/domain_inference.py
```diff
@@ -0,0 +1,223 @@
...
+    return use_domains
+
+
+def _infer_domain_from_extents(output_domain: sast.Cartesian,
+                               extents: sast.Extent,
+                               intervals: Sequence[sast.Interval]) -> sast.Cartesian:
+    """
+    Given the output domain and extents, infers the domain size of the input field.
+    Assuming that the output is accessed at offset (0, 0, 0).
+    # TODO Adapt for non-zero offsets
```
tbennun: Is this TODO not resolved? Also, why is it in the docstring? (additionally, please fix the docstring's formatting, i.e., document return value and add newline)

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+        node.operation_type.destination[0].extent.extents[:] = self._offsets_of_uses_in_scope(computation, node.result)
+        node.operation_type.destination[0].extent.sort_extents()
+
+    def _offsets_of_uses_in_scope(self, computation: sast.ComputationBlock, identifier: sast.Identifier):
+        """
+        Get the offsets of the uses of a field in the current scope.
+        :param computation: The current computation block
+        :param identifier: The identifier
+        :return: A list of offsets
+        """
```
tbennun:
```suggestion
        """
        Get the offsets of the uses of a field in the current scope.

        :param computation: The current computation block
        :param identifier: The identifier
        :return: A list of offsets
        """
```

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+
+        return offsets
+
+
+def _init_outputs(dtypes: Sequence[sast.FieldType]):
+    """
+    Initialize the extents of every element to be (0, 0, 0)
+    :param dtypes:
+    :return:
+    """
```
tbennun:
```suggestion
    """
    Initialize the extents of every element to be (0, 0, 0)

    :param dtypes: Field type elements to set.
    """
```

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _minkowski_sum(a: Collection[sast.Offset],
+                   b: Collection[sast.Offset]) -> set[sast.Offset]:
+    """
+    Compute the minkowski sum of two sets of extents
+    :param a:
+    :param b:
+    :return:
+    """
```
tbennun:
```suggestion
    """
    Compute the Minkowski sum of two sets of extents.

    :param a: First set operand.
    :param b: Second set operand.
    :return: The Minkowski sum set result.
    """
```

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+        # Concatenate all the extents from uses_of_result
+        offsets = []
+        for use in uses_of_result:
+            offsets.extend(use.field_type.extent.extents)
+        if len(offsets) == 0:
+            print(f"WARNING: No uses of {identifier.name} found within current scope.")
+
+        # Remove unknown offsets (TODO: Investigate why this can happen!)
+        offsets = [o for o in offsets if not o.is_unknown()]
+        assert all(not o.is_unknown() for o in offsets), f"Offsets of uses of {identifier.name} must be known before inference"
```
tbennun: Don't you want to move the assertion before the filtering? Also, when does this happen?

spatialstencil/syntax/stencil_ir/extent_inference.py
```diff
@@ -0,0 +1,195 @@
...
+    Then, we compute the extents for each field based on the uses,
+    by traversing the program backwards and accumulating the extents.
+    Specifically, the extent of an input to a computation or program is the union of all its uses.
+    The extent of a statement argument is the Minkowski sum of its 'local' extent and the extents of its uses.
+
+    The 'local' extents are computed using an ExtentCollector.
+
+    :param program:
+    :return:
+    """
```
tbennun:
```suggestion
    :param program: The program whose extents should be inferred (operates in-place).
    """
```

spatialstencil/syntax/stencil_ir/domain_inference.py
```diff
@@ -0,0 +1,223 @@
...
+def _domains_of_uses_in_scope(uses: dict[sast.Identifier, Collection[def_use_analysis.ScopedUse]],
+                              computation: sast.ComputationBlock | sast.Program,
+                              identifier: sast.Identifier) -> list[sast.Cartesian]:
+    """
+    Get the offsets of the uses of a field in the current scope.
+    :param uses: The uses dictionary
+    :param computation: The current computation block
+    :param identifier: The identifier
+    :return: A list of offsets
+    """
```
tbennun:
```suggestion
    """
    Get the offsets of the uses of a field in the current scope.

    :param uses: The uses dictionary
    :param computation: The current computation block
    :param identifier: The identifier
    :return: A list of offsets
    """
```

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -323,6 +548,8 @@ def as_ir(self, indent: int = 0) -> str:
 class MathCall(Node):
     """
     A mathematical function call operation for stateless calls
+
+    # TODO: Why do math calls not have types?!
```
tbennun: A math call is not an operation, so it does not have a type. Their type is inferred as part of the operation they partake in.
It would be the same as asking "why does a unary operator IR node not have a type".

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -449,6 +687,8 @@ def as_ir(self, indent: int = 0) -> str:
 class ElseIfBlock(Node, Block):
     """
     A single "else if" block. If condition is None, represents an "else" block
+
+    # TODO: Does not have an operation type!!
     """
```
tbennun: An else-if block is always nested in an IfBlock, which does have one. It is important to fix the IfBlock (which is the Operation here) type and the other blocks to have the same operation type.
Notice that ElseIfBlock is _not_ an Operation in and of itself.

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -566,6 +825,57 @@ def __init__(self, *args, **kwargs):
...
+        :param computation:
+        :return:
+        """
+        self.generic_visit(computation)
+
+    def post_visit_ComputationBlock(self, computation: ComputationBlock):
+        """
+        Called after leaving the scope of the computation block.
+        :param computation:
+        :return:
```
tbennun: Please fix docstring formatting (newline before parameter, and if there is no documentation for the params or return value, they should be removed).

tbennun (changes requested):
Overall good changes, but domain inference seems to still have a bug with non-materialized codes and we should discuss some aspects of the SSA definition w.r.t. scope isolation.

samples/spst/laplacian_no_mat_ext_dom.spst
luxfg: No, the test is failing on purpose. The important thing is that we should define it so that the input field %in has the same size regardless of if you materialize or not. Moreover, one needs to interpret a temporary field with more than one extent as having at each coordinate not just its own gridpoint value, but that of all extents. Let's walk through the example (it's easier to verify bottom up).


```
        // We need to compute for each gridpoint in [0:128, 0:128, 0:80] its value and the value of its 0,1,0 neighbor. For the first, we need %in domain [-1:129, -1:129, 0:80], for the second %in domain [-1:129, 0:130, 0:80]. The result is the union of the two domains.
        %out_1 = spst.statement(%in)
                : field<[-1:129,-1:130,0:80], {(0, 0, 0), (-1, 0, 0), (1, 0, 0), (0, -1, 0), (0, 1, 0), (0, 1, 0), (-1, 1, 0), (1, 1, 0), (0, 2, 0)}, f32>
                -> field<[0:128, 0:128, 0:80], {(0, 0, 0), (0, 1, 0)}, f32>
            {
                return -4.0 * %in[0, 0, 0] + %in[-1, 0, 0] + %in[1, 0, 0] + %in[0, -1, 0] + %in[0, 1, 0]
            }
         // Since we have for each gridpoint in %out_1 its value and the value of 0,1,0, and no other values are needed for the 0,0,0 offset, the domain size of %out_1 matches that of %out_2
        %out_2 = spst.statement(%out_1)
                : field<[0:128, 0:128, 0:80], {(0, 0, 0), (0, 1, 0)}, f32>
                -> field<[0:128, 0:128, 0:80], {(0, 0, 0)}, f32>
            {
                return %out_1[0, 0, 0] + %out_1[0, 1, 0]
            }
         // The output size is given
        return %out_2 : field<[0:128, 0:128, 0:80], {(0, 0, 0)}, f32>
```

spatialstencil/lowering/gt4py_to_stencil_ir.py
luxfg: I mean the cleanest is to have a pass on the stencil ir that transforms any program into SSA.

spatialstencil/syntax/stencil_ir/irnodes.py
luxfg: Yes, makes sense.

spatialstencil/syntax/stencil_ir/irnodes.py
luxfg: Yes, makes sense.

spatialstencil/syntax/stencil_ir/irnodes.py
luxfg: sure

spatialstencil/syntax/stencil_ir/def_use_analysis.py
luxfg: Yes, this should be done elsewhere.It fails if you don't explicitly annotate all the return types as field types (the hdiff, uvbke, vadv, hdiffsa examples don't annotate returns yet for example). We should have a pass that fills in 'any' types with either scalar or field types depending on the context.

spatialstencil/syntax/stencil_ir/lark_to_ast.py
luxfg: Arguments can have different types. For example. you can have computation that returns an f32 fields and an f16 field. Moreover, some fields could be "2D" while others are "3D". So you cannot rely on the last value only.

spatialstencil/lowering/gt4py_to_stencil_ir.py
luxfg: I think we should have new versions in the if-else to ensure it's strictly SSA. It simplifies the use-def / def-use analysis.

spatialstencil/syntax/stencil_ir/def_use_analysis.py
luxfg: Yeah in principle, computations and statements (and programs) are indeed 'isolated from above' scopes. We are not using or checking that here.

luxfg: It might be nice to actually check the isolation of scopes.

luxfg: I understand now what you meant. As discussed, we will only enforce new versions within the isolatedFromAbove scopes.

luxfg:
Ok, I had another look at some vertical stencil examples and writing to input arrays, and thought about the view vs field concepts. Here is my current understanding:

---

fields vs views:

- computations should operate on views, because when we have intervals they access only part of the field. In terms of lowering to spatial, the views of arguments should convey the bounds of the computation. The vertical stencil example (vertical_intervals_ext_dom.spst) clarifies this idea.
- programs should operate on fields.

I realize that the computations still have ? argument extent types, which is a bit awkward, but conceptually they have to be views I think. If we wanted to be super clear, we could have a subtype of views that do not have extents and are exclusively used for compute arguments. The benefit of this is marginal I think though.

What do you think?

Can you adjust the lowering to create FieldType / ViewType appropriately?

----

Writing to computation arguments:

- when we write to an input array view, we should return the view that we overwrite and assign it to the corresponding field. This is the cleanest way to ensure proper def-use chains without creating cycles and is the clearest semantically.
- Using arguments as silently being used by statement that write to them creates cycles in the analysis, which is problematic and would require an iterative method I think. Using explicit returns keeps the analysis 1 pass. Hence, I strongly encourage to use the return approach.

I have made an example  (vertical_readwrite[_ext_dom].spst) to demonstrate this interpretation. 

Does that make sense?

If you agree and we don't find any important cases this does not cover, can you adjust the codegen to generate proper returns?

luxfg:
I adjusted the lowering to use fields in the program arguments and fixed the SSA versioning for nested if-else, by implementing a proper SSA pass in the stencil IR. The only remaining issue is the proper creation of inputs/outputs. The cleanest solution would be to use the Def-Use / Use-Def chains, but it might also be possible to extend the current input/output visitor.

luxfg:
I adjusted the output pruning to keep outputs that are used as arguments to SOME computation. This is a conservative way to ensure that we do not remove fields that are over-written.

tests/test_lowering_gt4py_to_stencil_ir.py
```diff
@@ -92,14 +92,15 @@ def test_lower_gt4py_intermediates_4(self):
...
-        assert [out.name for out in comps[1].outputs] == ['out']
+        assert [out.name for out in comps[1].outputs] == ['out', 'tmp']
         extents = analysis.collect_extents(comps[0])
         assert next(iter(extents['inp'].values())) == {(0, -1, 0)}
         assert next(iter(extents['tmp'].values())) == {(1, 0, 0), (0, 0, 0)}
 
     def test_lower_gt4py_intermediates_5(self):
         program = self.gtfuncs['intermediates_versioning_3']
         irprogram = gt4py_to_stencil_ir.lower_gt4py_to_stencil_ir(program, materialize=False)
+        print(irprogram.as_ir())
```
tbennun:
```suggestion
```

tests/test_lowering_gt4py_to_stencil_ir.py
```diff
@@ -112,37 +113,39 @@ def test_lower_gt4py_intermediates_5(self):
...
+        assert comps[0].outputs == [sast.Identifier('tmp', version=0)]
+        assert comps[1].inputs == [sast.Identifier('tmp', version=0)]
+        assert comps[1].outputs == [sast.Identifier('out', version=0), sast.Identifier('tmp', version=1)]
 
     def test_versioning_in_ifelse(self):
         program = self.gtfuncs['versioning_in_ifelse']
         irprogram = gt4py_to_stencil_ir.lower_gt4py_to_stencil_ir(program, materialize=False)
 
+        print(irprogram.as_ir())
+
```
tbennun:
```suggestion
```

tests/test_lowering_gt4py_to_stencil_ir.py
```diff
@@ -161,12 +164,13 @@ def test_output_overwrite(self):
     def test_lower_gt4py_if(self):
         program = self.gtfuncs['satadjust_specific_humidity']
         irprogram = gt4py_to_stencil_ir.lower_gt4py_to_stencil_ir(program)
+        print(irprogram.as_ir())
```
tbennun:
```suggestion
```

tests/test_stencil_ir_passes.py
```diff
@@ -144,6 +171,7 @@ def test_infer_extents_and_domains_is_complete(self):
         files = [
             Path(__file__).parent / Path('../samples/spst/hdiff.spst'),
             Path(__file__).parent / Path('../samples/spst/vadv.spst'),
+            #Path(__file__).parent / Path('../samples/spst/vadv_3.spst'),
```
tbennun:
```suggestion
```

tbennun (approved):
Just a little bit of cleanup, otherwise looks good!

PR #11, opened 2024-09-19
Spatial IR Syntax Implementation

luxfg:
Changes to syntax/spatial

* Draft of the Spatial IR Syntax Implementation. 

Note that it is a fairly direct translation of the syntax and does not include any control flow graph notions (yet).

Changes to syntax/common

* Refactor the scalar types into common, so we can reuse them between IRs.
* validate() should be part of the base node.
* I fixed validate_schema to work properly with UnionTypes using Union or | syntax.
* validate_schema is a class method.

Changes in syntax/stencil

* call validate_schema in the Program validate()

### DIFF
--- a/spatialstencil/syntax/common/basenode.py
+++ b/spatialstencil/syntax/common/basenode.py
@@ -82,8 +82,6 @@ class BaseNode:
         for field_name, field in cls.__dataclass_fields__.items():
             # Resolve the field's type using get_type_hints (handling forward references)
             field_type = type_hints[field_name]
-            print(field_type)
-
             origin = typing.get_origin(field_type)
             if origin is types.UnionType or origin is typing.Union:
                 _check_union(field_type, field_name)
--- a/spatialstencil/syntax/common/basenode.py
+++ b/spatialstencil/syntax/common/basenode.py
@@ -33,7 +33,7 @@ class BaseNode:
     """
 
     @classmethod
-    def validate_schema(cls, visited: set[type['BaseNode']] = None) -> bool:
+    def validate_schema(cls, visited: set[type['BaseNode']] = None):
         """
         Validates that the node type and all its child node types abide by
         the rules defined on ``BaseNode``.
@@ -94,7 +94,6 @@ class BaseNode:
             else:
                 if not isinstance(field_type, type) or not issubclass(field_type, (int, float, str, type(None), Enum)):
                     raise TypeError(f'Unsupported terminator type {field_type} for field {field_name} of {cls}')
-        return True
 
     def validate(self) -> None:
         """
--- a/spatialstencil/syntax/common/basenode.py
+++ b/spatialstencil/syntax/common/basenode.py
@@ -3,6 +3,7 @@ Provides a base class for AST/IR trees. Includes children queries and schema val
 """
 import types
 import typing
+import warnings
 from dataclasses import dataclass
 from collections import deque
 import pprint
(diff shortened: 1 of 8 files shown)


spatialstencil/syntax/common/basenode.py
```diff
@@ -29,7 +33,7 @@ class BaseNode:
     """
 
     @classmethod
-    def validate_schema(cls, visited: set[type['BaseNode']] = None) -> None:
+    def validate_schema(cls, visited: set[type['BaseNode']] = None) -> bool:
```
tbennun: does it return true, false, and raise exceptions? I'd go with one or the other (also see/update docstring below as necessary)

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -0,0 +1,571 @@
...
+    A range expression (start:stop or start:stop:step).
+    """
+    start: Expression
+    stop: Expression
+    step: Expression = None
+
+    def validate(self) -> None:
+        assert isinstance(self.start, Expression)
+        assert isinstance(self.stop, Expression)
+        if self.step:
```
tbennun:
```suggestion
        if self.step is not None:
```
if self.step were set to 0 directly it would not enter the condition

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -0,0 +1,571 @@
...
+    def validate(self) -> None:
+        if isinstance(self.hops, list):
+            for dx, dy in self.hops:
+                assert abs(dx) + abs(dy) == 1, "Each hop must have an absolute sum of 1."
+
+    def as_ir(self, indent: int = 0) -> str:
+        indent_str = '  ' * indent
+        hops_str = "auto" if self.hops == "auto" else f"[{', '.join(hop.as_ir() for hop in self.hops)}]"
+        channel_str = "auto" if self.channel == "auto" else str(self.channel)
+        return f"{indent_str}hops = {hops_str}, \n{indent_str}channel = {channel_str}"
```
tbennun:
```suggestion
        return f"{indent_str}hops = {hops_str},\n{indent_str}channel = {channel_str}"
```

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -0,0 +1,571 @@
...
+    """
+    A routing declaration for a stream, optionally specifying hops and channel.
+    """
+    hops: Union[list[RoutingHop], Literal["auto"]] = "auto"  # list of hops or 'auto'
+    channel: Union[int, Literal["auto"]] = "auto"  # Channel ID or 'auto'
+
+    def validate(self) -> None:
+        if isinstance(self.hops, list):
+            for dx, dy in self.hops:
+                assert abs(dx) + abs(dy) == 1, "Each hop must have an absolute sum of 1."
```
tbennun, line 254: nice semantics validation

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -0,0 +1,571 @@
...
+    name: str | None
+    parameters: list[Parameter]
+    arguments: list[KernelArgument]
+    body: list[PlaceBlock | DataflowBlock | ComputeBlock | Phase]
+
+    def validate(self) -> None:
+        assert all(isinstance(p, Parameter) for p in self.parameters)
+        assert all(isinstance(arg, KernelArgument) for arg in self.arguments)
+        assert all(isinstance(stmt, (Phase, ComputeBlock, DataflowBlock, PlaceBlock)) for stmt in self.body)
+        assert self.validate_schema()
```
tbennun: is this necessary inside node validation? normally the schema is validated separately.

spatialstencil/syntax/stencil_ir/irnodes.py
```diff
@@ -836,6 +788,7 @@ def validate(self) -> None:
         assert all(isinstance(i, (FieldType, ScalarType, AnyType)) for i in self.operation_type.source)
         assert all(isinstance(i, (FieldType, ScalarType, AnyType)) for i in self.operation_type.destination)
         assert isinstance(self.computations[-1], ReturnOp)
+        assert self.validate_schema()
```
tbennun: Same question here. Perhaps validate_schema on each of the IRs should be a test (part of CI)

tbennun (changes requested):
Looks good, very minor changes requested

spatialstencil/syntax/stencil_ir/irnodes.py
luxfg: Yes, I think you are right.

tests/test_spatial_ir_schema.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+
+from spatialstencil.syntax.spatial_ir.irnodes import Kernel
+
+
+class TestStencilIR(unittest.TestCase):
```
tbennun:
```suggestion
class TestSpatialIR(unittest.TestCase):
```

tbennun (approved):

PR #13, opened 2024-09-27
Spatial IR Harmonization and Parser

tbennun:
Adds a Lark/AST parser for Spatial IR. Also makes the following modifications to the standard in order to harmonize the representation:

- [x] Blocks define completion or await, fail when neither exist
- [x] Assignment of a completion object is disallowed
- [x] Typed identifiers become part of the block
- [x] `foreach` accepts generators that are implicitly zipped together. `receive(x)` is the only built-in generator and is required
- [x] `receive(arr, stream)` is a built-in function that can be used as a shorthand for `foreach i32 i, var x in [0:N], receive(stream) { arr[i] = x; }`
- [x] Parametric place/dataflow/compute block indices also have a data type

### DIFF
--- a/spatialstencil/syntax/spatial_ir/irnodes.py
+++ b/spatialstencil/syntax/spatial_ir/irnodes.py
@@ -101,13 +101,10 @@ class TypedIdentifier(SpatialNode):
     A variable identifier (e.g., x, y, my_variable) with a type.
     """
     dtype: Union[ScalarType, StreamType, ArrayType]
-    name: str
-    version: int
+    identifier: Identifier
 
     def as_ir(self, indent: int = 0) -> str:
-        if self.version == 0:
-            return f'{self.dtype.as_ir()} {self.name}'
-        return f'{self.dtype.as_ir()} {self.name}#{self.version}'
+        return f'{self.dtype.as_ir()} {self.identifier.as_ir()}'
 
 
 # Unary Operators
@@ -426,7 +423,7 @@ class ForeachStatement(Statement):
     variables: list[TypedIdentifier]
     parameter_range: list[RangeExpression]
     stream_variable: Identifier
-    receive_stream: Identifier
+    receive_stream: ReceiveGenerator
     body: list[Statement]
     completion_name: Optional[Completion] = None
 
@@ -440,9 +437,9 @@ class ForeachStatement(Statement):
         body_str = "\n".join(stmt.as_ir(indent + 1) for stmt in self.body)
 
         if self.parameter_range:
-            main_str = f'foreach {vars_str} in [{rng_str}], receive({self.receive_stream.as_ir()}) {{\n{body_str}\n{indent_str}}}'
+            main_str = f'foreach {vars_str} in [{rng_str}], {self.receive_stream.as_ir()} {{\n{body_str}\n{indent_str}}}'
         else:
-            main_str = f'foreach {vars_str} in receive({self.receive_stream.as_ir()}) {{\n{body_str}\n{indent_str}}}'
+            main_str = f'foreach {vars_str} in {self.receive_stream.as_ir()} {{\n{body_str}\n{indent_str}}}'
 
         if self.completion_name:
             return f'{indent_str}{self.completion_name.as_ir()} = {main_str}'
@@ -536,25 +533,6 @@ class AssignmentStatement(Statement):
     def as_ir(self, indent: int = 0) -> str:
         indent_str = '  ' * indent
         return f'{indent_str}{self.destination.as_ir()} = {self.source.as_ir()}'
-
-
-@dataclass
-class DefinitionStatement(Statement):
-    """
-    Assigns the result of an expression to a field or variable
-    """
-    dtype: Union[ScalarType, StreamType]
-    destination: Identifier
-    source: Expression
-
-    def validate(self) -> None:
-        assert isinstance(self.source, Expression)
-        assert isinstance(self.destination, (ArraySlice, Identifier))
-        assert isinstance(self.dtype, (ScalarType, StreamType))
-
-    def as_ir(self, indent: int = 0) -> str:
-        indent_str = '  ' * indent
-        return f'{indent_str}{self.dtype.as_ir()} {self.destination.as_ir()} = {self.source.as_ir()}'
 
 
 # Compute Block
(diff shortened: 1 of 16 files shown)


spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -369,21 +423,26 @@ class ForeachStatement(Statement):
     """
     Foreach loop for asynchronously iterating over a received stream.
     """
-    variables: list[Identifier]
-    receive_stream: Receive
+    variables: list[TypedIdentifier]
+    parameter_range: list[RangeExpression]
+    stream_variable: Identifier
+    receive_stream: Identifier
```
luxfg: Why not use the receiveGenerator here?

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -474,18 +538,37 @@ def as_ir(self, indent: int = 0) -> str:
         return f'{indent_str}{self.destination.as_ir()} = {self.source.as_ir()}'
 
 
+@dataclass
+class DefinitionStatement(Statement):
```
luxfg: Is it an assignment or a definition?

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -72,18 +83,33 @@ class ArrayType(SpatialNode, IRType):
     An array type of a scalar or stream, with one or more dimensions.
     """
     base_type: Union[ScalarType, StreamType]
-    shape: list[Union[int, Parameter]]
+    shape: list[Union[int, 'Expression']]
 
     def validate(self) -> None:
         assert isinstance(self.shape, list)
-        assert all(isinstance(dim, (int, Parameter)) for dim in self.shape)
+        assert all(isinstance(dim, (int, Expression)) for dim in self.shape)
```
luxfg, line 106: Ok. We should probably add a check that the Expression is compile-time known.

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -266,7 +305,7 @@ class RelativeStreamDeclaration(SpatialNode):
     A stream declaration inside a dataflow block that declares a communication stream
     to and from PEs at relative positions, with an optional routing declaration.
     """
-    dtype: StreamType
+    dtype: ScalarType
```
luxfg: Why is the type a scalar?

spatialstencil/syntax/spatial_ir/irnodes.py
luxfg: I am not sure where this is used. It's not possible to define a field in a statement. It has to happen in a field declaration within the dataflow block. and for assignments, use the AssignmentStatement.

spatialstencil/syntax/spatial_ir/language.lark
```diff
@@ -0,0 +1,203 @@
...
+
+for_stmt : "for" typed_vars "in" subgrid_expression compute_body
+map_stmt : prefix "map" typed_vars "in" subgrid_expression compute_body
+foreach_stmt : prefix "foreach" typed_vars "in" generators compute_body
+async_stmt : completion "async" compute_body
+
+// Top-level statements that are direct children of `compute` blocks
+?base_stmt : (function_call | await_completion | map_stmt | foreach_stmt | async_stmt | for_stmt)
+
+definition : builtin_type identifier "=" value_expr
```
luxfg: you cannot define an array or other field in a statement.

statements like

T[80] a = ...
f32 a = 5

in the compute block are not allowed.
all data must be declared in the place block using a field declarations.

am I missing sth?

spatialstencil/syntax/spatial_ir/irnodes.py
luxfg: (not in this PR though)

spatialstencil/syntax/spatial_ir/irnodes.py
```diff
@@ -72,18 +83,33 @@ class ArrayType(SpatialNode, IRType):
...
         return f'{self.base_type.as_ir()}[{dims}]'
 
 
+@dataclass
+class TypedIdentifier(SpatialNode):
+    """
+    A variable identifier (e.g., x, y, my_variable) with a type.
+    """
+    dtype: Union[ScalarType, StreamType, ArrayType]
+    name: str
```
luxfg: Consider having

dtype: ...
identifier: Identifier

as a signature.

luxfg: this allows tracking identifiers regardless of if they are typed or not.

luxfg (changes requested):
Nice overall. I am not sure what the DefinitionStatement is doing.

spatialstencil/syntax/spatial_ir/irnodes.py
tbennun: Based on the types stream can accept, they are a scalar type:
A `stream<completion>` cannot exist, nor can a `stream<f32[N]>`

spatialstencil/syntax/spatial_ir/irnodes.py
tbennun: True, this is checked in lowering to CSL (actually, it cannot be parametric there either).

spatialstencil/syntax/spatial_ir/irnodes.py
tbennun: I was trying to keep things simpler, but we can use the generator instead. Will fix

spatialstencil/syntax/spatial_ir/language.lark
tbennun: you're right. Will fix. This also removes the `DefinitionStatement` above.

spatialstencil/syntax/spatial_ir/irnodes.py
tbennun: Correct, will fix!

spatialstencil/syntax/spatial_ir/irnodes.py
luxfg: The StreamType is defined such that it can only hold scalar elements, so your example cannot occur in either case.

```
class StreamType(SpatialNode, IRType):
    """
    A stream type that sends elements of type T.
    """
    element_type: ScalarType
```

My interpretation is that the stream declaration creates an object that is a stream, which is why its dtype should be StreamType. Yes, the stream type holds a scalar, and it is accessible through the element_type of the stream.

luxfg (approved):
Ok. I changed the type of stream declaration as per my reasoning in the comment. I added type validations to every irnode and fixed some small issues in the type annotations and lark_to_ir.

PR #24, opened 2025-10-02
Add benchmarking support

tbennun:
Adds cycle count benchmark handlers in the code. Benchmarking functionality is only enabled if `--benchmark` is given to the program. Otherwise, program performance is not affected (except for a ~12-byte memory overhead per PE, which can be disabled at compile-time).


luxfg:
I recall that Piotr had some fairly sophisticated set-up that would use a bcast/reduce to sync up the PEs and then use this to get more reliable cycle counts. I am not sure though if this is needed for our setting, just wanted to know/discuss if you purposefully decided to use a different set up or have any other experience points regarding benchmarking WSE.

luxfg (approved):
looks good, but let's discuss the benchmarking methodology.

PR #25, opened 2025-10-03
Support strided place/dataflow/compute blocks

tbennun:


spatialstencil/syntax/spatial_ir/grid_geometry.py
```diff
@@ -64,7 +66,7 @@ def __str__(self):
 # RECTANGLE SPLITTING
 ###
 
-def _ranges_overlap(range1: tuple[int, int], range2: tuple[int, int]) -> bool:
+def _ranges_overlap(range1: tuple[int, int, int], range2: tuple[int, int, int]) -> bool:
```
luxfg: To actually answer this we would need a different kind of logic. Your solution does not work in the general case, maybe we should make a note/add an assertion that it assumes the stride is at most 2? in general I think the solution involves modular arithmetic.

luxfg (changes requested):
looks good! I'd at least comment on or better add assertion on the limitations of the grid intersection algo.

spatialstencil/syntax/spatial_ir/grid_geometry.py
luxfg: I am not 100% sure it is correct for the stride 1/2 case either, like why are you doing the min of the strides? it seems like if we have stride 1 and stride 2, the overlap would either have stride 2 or not exist.
Also I think there needs to be some accounting for the start/stop having same parity or it cannot work.

Where do we need this? If not actually needed anywhere in the pipeline, we can add assertion for stride =1 / or raise unsupported error. Otherwise, needs to be fixed I think.

luxfg:
@glukas fix tomorrow

luxfg (approved):
I have generalized the grid geometry to handle the strided case as well.

It was a bit more tricky than I thought, but I think it's ok for our purposes now.

It assumes x-strides are equal, and y-strides are equal. 

If they are not because one of them has stride 1, you can apply a checkerboarding transformation first. should be fine for everything we care about for now.

let me know if we need to handle a more general case.

luxfg:
note that the failing test is the one we disabled in main

PR #32, opened 2025-10-29
Copy elimination pass

tbennun:
Blocked by #33

The pass is written as two separate steps: Extraneous copy detection and pruning unused variables from `place` blocks.
- [x] Prune unused fields + Tests
- [x] Eliminate extraneous copies + Tests
- [x] Allow `send` with constant values + Tests
- [x] Make elision and pruning on by default


spatialstencil/syntax/spatial_ir/passes.py
luxfg, line 1: I think it's starting to be of a size where it makes sense to split into multiple files according to pass. Maybe a passes folder with individual passes.

spatialstencil/syntax/spatial_ir/passes.py
```diff
@@ -152,50 +157,100 @@ def constexpr_propagation(kernel: spa.Kernel) -> spa.Kernel:
...
+    def visit_DataflowBlock(self, node: spa.DataflowBlock):
+        # Do not traverse into dataflow blocks
+        return node
+
+    def visit_Identifier(self, node: spa.Identifier):
+        self.used.add(node)
+        return node
+
+
+def prune_unused_fields(kernel: spa.Kernel) -> spa.Kernel:
```
luxfg: Very nice and clean!

spatialstencil/syntax/spatial_ir/passes.py
luxfg: #maybe make an __ init __ for the passes module that exports the passes as functions?

spatialstencil/syntax/spatial_ir/passes.py
```diff
@@ -222,9 +277,709 @@ def _regisiter_read(self, s: spa.ReceiveGenerator | spa.ReceiveGenerator):
...
+    if src_access is None:
+        return None
+    if src_access[1] is not None and dest_access[1] != src_access[1]:
+        return None
+    if src_access[1] is None and dest_access[1]:  # Set scalar into array index
+        return None
+    return CopyCandidate(dest_access[0], src_access[0])
+
+
+def is_copy(statement: spa.AssignmentStatement | spa.MapStatement) -> CopyCandidate | None:
```
luxfg: We could use the treematching.PatternMatcher functionality. Note sure if all the logic can be handled, but at least the structural matching can be done using that. These multiple if/else logic can be a bit error prone. Just a thought.

spatialstencil/syntax/spatial_ir/passes.py
```diff
@@ -222,9 +277,709 @@ def _regisiter_read(self, s: spa.ReceiveGenerator | spa.ReceiveGenerator):
...
+
+    if hasattr(statement, "body") and isinstance(getattr(statement, "body"), list):
+        header_stub = replace(statement, body=[])
+        reads, writes = _collect_reads_writes(header_stub)
+    else:
+        reads, writes = _collect_reads_writes(statement)
+    return set(reads), set(writes)
+
+
+class _LivenessAnalyzer:
```
luxfg: If we wanted to build more robust infrastructure, we should extract the Control Flow Graph and then run the algorithm on the more abstract structure. The benefit is that we will need an explicit CFG representation for other passes, in particular coloring as well. And so rather than hand crafting this logic each time we could build on a shared representation.

spatialstencil/syntax/spatial_ir/passes.py
```diff
@@ -222,9 +277,709 @@ def _regisiter_read(self, s: spa.ReceiveGenerator | spa.ReceiveGenerator):
...
+    """Backward dataflow analysis computing per-statement liveness sets."""
+
+    def __init__(self):
+        self.records: dict[int, _LivenessRecord] = {}
+
+    def run(
```
luxfg: The standard algorithm I know is a fixed point algorithm. Are we iterating somewhere else? Or can we shortcut this due to our particular structure?

luxfg: I see that you iterate only inside of the loop. Could you show why this is still correct?

spatialstencil/syntax/spatial_ir/passes.py
```diff
@@ -222,9 +277,709 @@ def _regisiter_read(self, s: spa.ReceiveGenerator | spa.ReceiveGenerator):
...
+        header_reads, header_writes = _direct_reads_writes(statement)
+        body_in = self._analyze_sequence(statement.body, set(live_out))
+        return header_reads | ((body_in | live_out) - header_writes)
+
+    def _analyze_loop(
+        self,
+        statement: spa.Statement,
+        live_out: set[spa.Identifier],
+    ) -> set[spa.Identifier]:
+        header_reads, header_writes = _direct_reads_writes(statement)
```
luxfg: Please explain algorithm and how it relates to the well known one. Standard iterates until two iterations do not change, here you break immediately.

tests/spatial_ir/test_spatial_ir_passes.py
```diff
@@ -1,6 +1,4 @@
-import unittest
-
-from spatialstencil.syntax.spatial_ir import irnodes as spir, canonicalization
+from spatialstencil.syntax.spatial_ir import irnodes as spir, canonicalization, passes, parser
```
luxfg, line 3: Add unit tests for liveness analysis.

luxfg (changes requested):

tests/spatial_ir/test_spatial_ir_passes.py
tbennun: no more liveness analysis in the codebase, so no need to test

spatialstencil/syntax/spatial_ir/copy_elimination.py
luxfg, line 1: Consider adding some documentation, for example like this:

### Sub-pass 1 — `RemoveRedundantCopies` (`_optimize_region`)

Runs on the compute block's flat statement list. It repeatedly scans for a **producer** that can be forwarded directly to its single consumer, then removes the producer. The loop restarts after each successful elimination.

Three producer shapes are recognised. In all cases the destination field must be written by exactly one statement and read by exactly one statement in the region — where "one statement" means one *distinct statement*, not one *access*, due to the binary counting in `_statement_field_counts`.

| Shape | Example | Matched by |
|---|---|---|
| Scalar copy | `tmp = a` | `_extract_direct_producer` |
| Element-wise map copy | `map(tmp[i,j] = a[i,j])` | `_extract_map_producer` |
| Foreach-receive into local buffer followed by a send | `foreach(src) { tmp[i] = x }` … `send(tmp, stream)` | `_extract_foreach_bulk_producer` |

For each producer the pass verifies:
- The destination is not *protected* (referenced by sibling statements outside the current nesting scope, to prevent incorrectly removing a field still live elsewhere).
- Nothing overwrites the producer's source between the producer and the consumer.

For the foreach-bulk shape the temporary intermediate destination (`tmp`) must be a **non-extern** field — it is a local on-chip buffer. The **source** (the stream field) must be a place field, but may be extern. The optimisation therefore implicitly assumes that extern source fields are directly usable as a send source without a prior materialisation step.

After the while-loop converges, `_optimize_region` recurses into the bodies of nested `for`/`foreach`/`map`/`async` blocks.

---

### Sub-pass 2 — `RemoveSingleElementIndexCopies` (`_optimize_single_element_index_region`)

Runs inside loop bodies (`for`/`foreach`/`map`). It handles patterns like:

```
tmp[i] = a[i]
out[i] = tmp[i]
```

The conditions are stricter than sub-pass 1:
- `tmp` must be written exactly **once across the whole region** (at the array level, not per index). This means two writes to different indices — `tmp[0] = …` and `tmp[1] = …` — block both from being optimised, even though their consumers are disjoint.
- Every read of `tmp[i]` must use the **exact same index signature** and be statically rewritable.
- The source field must not be overwritten between the producer and the last consumer.

luxfg (approved):
I think it strikes a good balance between effectiveness and complexity! Nice test suite. Maybe add some docs on the algorithm.

(review AI proposal below...)

spatialstencil/syntax/spatial_ir/copy_elimination.py
tbennun: Done

PR #33, opened 2025-10-29
Lower arguments to external streams and fields

tbennun:
During lowering to hardware, arguments are lowered into explicit `place` fields or `extern_stream`s depending on the mode of operation.

 In memcpy mode:
```rust
place ... {
   extern f16[K] inp;
}
```
In streaming mode:
```rust
dataflow ... {
  stream<f16, K> inp = extern_stream(in) { /*routing data*/ };
  stream<f16, K> out = extern_stream(out) { /*routing data*/ };
}
```

- [x] Implement syntax and tree nodes for external streams and fields
- [x] Lowering pass
- [x] CSL code generation support
    - [x] `extern` --> `export`
    - [x] Proper DSD management
    - [x] Benchmarking infrastructure redone as extern fields
    - [x] Go over all uses of `kernel.arguments` for cleanup

### DIFF
--- a/spatialstencil/cli/compiler.py
+++ b/spatialstencil/cli/compiler.py
@@ -54,7 +54,9 @@ def compile_spatial_ir(input_file: str, output_folder: str, param: list[str], of
 
     # Argument checks
     using_memcpy_mode = None
+    argnames = []
     for arg in kernel.arguments:
+        argnames.append(arg.identifier.name)
         # Change all shapes to be lists of integers
         if isinstance(arg.dtype, spa.ArrayType):
             arg.dtype.shape = [dim.eval() if isinstance(dim, spa.Expression) else int(dim) for dim in arg.dtype.shape]
@@ -136,7 +138,7 @@ def compile_spatial_ir(input_file: str, output_folder: str, param: list[str], of
         "kernel_name": kernel.name,
         "inputs": input_args,
         "outputs": output_args,
-        "argument_order": [a.identifier.name for a in kernel.arguments],
+        "argument_order": argnames,
         "memcpy_mode": using_memcpy_mode,
         "fabric_dims": [xend - xbegin, yend - ybegin],
         "kernel_dims": kernel_dims,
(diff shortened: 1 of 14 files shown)


tests/spatial_ir/test_argument_lowering.py
luxfg, line 1: - add test where the input rectangle is not equal to the compute rectangle and shifted. For example, when we have halos, inputs are offset by the halo. 
- It might be ok to "over-allocate" the inputs and put the arguments into every PE (as you have done I think), since later passes (unused field elimitation) can remove them)
- But need to check that we never "under-allocate"

luxfg: However, if we do 'over-allocate', I wonder how the runtime will know which part of the arguments to copy to which PE. Do we store the offsets separately? I do not fully understand this yet.

spatialstencil/syntax/spatial_ir/canonicalization.py
```diff
@@ -434,14 +390,82 @@ def visit_ForeachStatement(self, node: spir.ForeachStatement):
...
     :param dtypes: A mapping of identifier to its type in the given rectangle.
-    :param kernel_arguments: The list of kernel arguments, used to determine which identifiers are arguments.
     """
-    rect.metadata.compute = _ForeachDataTaskToLoopConverter(dtypes, kernel_arguments).visit(rect.metadata.compute)
+    rect.metadata.compute = _ForeachDataTaskToLoopConverter(dtypes).visit(rect.metadata.compute)
+
+
+def lower_arguments_to_extern(rectangles: list[Rectangle[PEBlock]], kernel: spir.Kernel) -> None:
+    """
+    Lowers stream arguments to extern stream declarations in a dataflow or place block.
```
luxfg: extern field declaration in a place block or extern stream  declaration in a dataflow block

spatialstencil/cli/compiler.py
```diff
@@ -117,11 +119,19 @@ def compile_spatial_ir(input_file: str, output_folder: str, param: list[str], of
     rectangles = canonicalization.consolidate_rectangles_to_equivalence_classes(kernel)
     stream_extents = analysis.detect_stream_argument_extents(rectangles, kernel)
```
luxfg, line 120: I might have misunderstood, but my thought was that external fields should be designed so that they "solve" the mapping issue at the SpaDa layer, that is, after lowering to external fields, there is no more inference on which data goes where. Maybe this was a misunderstanding? We might have to discuss.

spatialstencil/syntax/spatial_ir/irnodes.py
luxfg, line 1: new IR is good.

luxfg (approved):
I think we should discuss the semantics of the external fields in more detail, we might have slightly different interpretations.

If it works for our use cases and is important for benchmarks, we can merge this PR  though and iterate on the design later.

PR #36, opened 2026-03-25
Add loop-based metaprogramming. Inlined early during compilation

tbennun:


luxfg (approved):
Very nice and clean PR! So powerful!!

PR #40, opened 2026-03-29
Add multicast support, improve examples

luxfg:
* Implements support for multicast as a SPADA language extension with a range expression as the stream offset:

```
        dataflow i16 i, i16 j in [0:PX, 0:PY] {
            stream<f32> bcast = relative_stream(0, [1:PY]) {
                hops = auto,
                channel = 0
            }
        }
        compute i16 i, i16 j in [0:PX, 0:1] {
            await send(x, bcast)
        }
        compute i16 i, i16 j in [0:PX, 1:PY] {
            await receive(x, bcast)
        }
```

* Adds CSL backend support for multicast
* Use multicast in examples (for gemv / matvec) -- provides a simpler and faster bcast phase


samples/spatial/blas/gemv.sptl
```diff
@@ -11,8 +11,7 @@
  * Algorithm:
  *   Phase 1: All PEs load their K×K block of A from the host.
  *   Phase 2: The j=0 column loads x; the i=0 column loads y from the host.
- *   Phase 3: Broadcast x in the Y direction (j=0 → j=PY-1) using a pipelined
- *            red/blue checkerboard — mirrors broadcast_1D.sptl on the j axis.
+ *   Phase 3: Multicast x in the Y direction (j=0 → j=PY-1) using native CSL multicast.
```
tbennun:
```suggestion
 *   Phase 3: Multicast x in the Y direction (j=0 → j=PY-1) using native hardware multicast.
```
Not CSL, but hardware IMO

samples/spatial/blas/gemv.sptl
tbennun, line 1: wow!

spatialstencil/lowering/spatial_ir_to_csl.py
```diff
@@ -1119,6 +1119,54 @@ def _collect_routes(rectangles: list[Rectangle[PEBlock]], color_maps: list[dict[
             if isinstance(stream.stream, spir.ExternStreamDeclaration):
                 continue  # Extern streams do not have on-chip routing
 
+            if isinstance(stream.stream, spir.MulticastStreamDeclaration):
```
tbennun: that's one version of multicast. Maybe MulticastRange vs. MulticastSet?

spatialstencil/lowering/spatial_ir_to_csl.py
```diff
@@ -1119,6 +1119,54 @@ def _collect_routes(rectangles: list[Rectangle[PEBlock]], color_maps: list[dict[
...
+
+                # Last receiver: deliver to RAMP only, no forwarding.
+                k_last = stop - 1
+                cx, cy = _coord(k_last)
+                routing_inst = INDENT + '@set_color_config(%s, %s, %s, .{ .routes = .{ .rx = .{%s}, .tx = .{RAMP} } });\n' % (
+                    cx, cy, color_name_outbound, rx_dir)
+                if routing_inst not in routing_instructions:
+                    inst += routing_inst
+                    routing_instructions.add(routing_inst)
+                continue
```
tbennun, line 1224: for later, I fear that we need more verification that multiple streams (within the same phase) don't overlap and give conflicting routing instructions.

spatialstencil/syntax/spatial_ir/canonicalization.py
```diff
@@ -436,6 +436,37 @@ def visit_RelativeStreamDeclaration(self, node: spir.RelativeStreamDeclaration):
         node.routing = new_routing
         return node
 
+    def visit_MulticastStreamDeclaration(self, node: spir.MulticastStreamDeclaration):
+        node = self.generic_visit(node)
+        # Inject default routing if none provided.
+        if node.routing is None:
+            node.routing = spir.RoutingDeclaration(hops=[], channel="auto")
+            return node
+        # Normalise hops: multicast does not use point-to-point hops.
```
tbennun:
```suggestion
        # Normalize hops: multicast does not use point-to-point hops.
```

spatialstencil/syntax/spatial_ir/canonicalization.py
```diff
@@ -436,6 +436,37 @@ def visit_RelativeStreamDeclaration(self, node: spir.RelativeStreamDeclaration):
...
+        new_routing = copy.copy(node.routing)
+        new_routing.hops = []
+        node.routing = new_routing
+        # Validate the range and fixed offset.
+        rng = node.multicast_range
+        start = rng.start.eval()
+        if start < 1:
+            raise ValueError(
+                f"Multicast stream range start must be >= 1, got {start}. "
+                "The sender is always at offset 0 and cannot be included in the receiver range."
```
tbennun: does that mean we can only multicast in one direction (positive?) and without an offset? We should document this somewhere (the SpatialNode docstring does not cut it)

tbennun (changes requested):
super minor changes. Very cool addition!!

samples/spatial/blas/gemv.sptl
luxfg: :)

spatialstencil/lowering/spatial_ir_to_csl.py
luxfg: Yeah, atm, the CSL compiler will complain if there are any direct conflicts, but it would be nice to have an analysis at the SPADA level.

spatialstencil/syntax/spatial_ir/canonicalization.py
luxfg: I added support for negative ranges and offset and documented it in the irspec. We still don't support ranges that are both negative and positive (could be added later).

PR #41, opened 2026-03-29
Add documentation for extern fields and streams

### LINKED ISSUE #34: Finalize and document external fields and streams
See discussion in #33

tbennun:
Closes #34


luxfg (approved):
reads nicely, thx.

luxfg:
What are your thoughts on Agent.md in source code dir? does it belong there? I think it can be useful, but must be kept up to date if so.

tbennun:
> What are your thoughts on Agent.md in source code dir? does it belong there? I think it can be useful, but must be kept up to date if so.

We can delete it. I usually add it to new repositories now. I will delete before merging

PR #45, opened 2026-04-05
Recycle overflowing task IDs with a state machine

tbennun:
Implements `TaskCreationBehavior.STATE_MACHINE_ON_OVERRUN`.

### DIFF
--- a/spatialstencil/syntax/csl/task_recycling.py
+++ b/spatialstencil/syntax/csl/task_recycling.py
@@ -1,0 +1,640 @@
+"""
+This module plans how logical CSL local tasks can share a smaller set of
+hardware local-task IDs when the program contains more local tasks than the
+target architecture exposes in :mod:`spatialstencil.syntax.csl.constants`.
+
+Terminology
+-----------
+
+Logical task
+        A ``tdag.CSLTask`` produced by ``create_csl_tasks``.  These are the tasks
+        the completion DAG wants to execute.
+
+Hardware slot
+        One reusable local-task ID from ``constants.LOCAL_TASK_IDS`` together with
+        one generated CSL task function.  The generated function acts as a
+        dispatcher for all logical tasks mapped to that slot.
+
+State
+        A dense integer assigned to one logical task within a slot.  Codegen emits
+        a slot-local state variable such as ``__task_slot_0_state`` and the shared
+        dispatcher selects the active logical task by branching on that value.
+
+Representative task
+        The first logical task assigned to a slot.  Codegen binds the generated CSL
+        task function once, using this task's logical alias.  All other logical
+        tasks in the same slot reuse the same hardware ID through additional
+        ``const task_<i>_id = @get_local_task_id(...)`` aliases.
+
+High-level methodology
+----------------------
+
+The planner works in three phases.
+
+1. Collect local tasks
+
+     Only ``task.task_type == 'local'`` participates in recycling.  Data tasks
+     have their own binding scheme and are not handled here.
+
+2. Decide whether recycling is needed
+
+     If the requested task-creation behavior forbids recycling, or if the number
+     of local tasks already fits in the available hardware IDs, the planner emits
+     a trivial one-task-per-slot mapping.
+
+3. Assign overflow tasks to slots
+
+    Each logical local task becomes a vertex in a conflict graph.  An edge means
+    two tasks must not share a slot.
+
+    By default the planner keeps the first ``len(constants.LOCAL_TASK_IDS)`` local
+    tasks on distinct hardware slots and only recycles the overflow tail.  That
+    preserves the early task structure and keeps each dispatcher as small as the
+    conflict constraints allow.  For comparison or experimentation, callers can
+    opt back into the older whole-graph coloring mode that recolors every local
+    task together.
+
+The important part is how the conflict graph is built.
+
+Safety criterion
+----------------
+
+Two logical tasks may share a hardware slot only if one of them is guaranteed
+to occur before *every possible trigger* of the other.
+
+That is stricter than simple topological order.  A plain topological order only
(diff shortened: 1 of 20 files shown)


spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+        A ``tdag.CSLTask`` produced by ``create_csl_tasks``.  These are the tasks
+        the completion DAG wants to execute.
+
+Hardware slot
+        One reusable local-task ID from ``constants.LOCAL_TASK_IDS`` together with
+        one generated CSL task function.  The generated function acts as a
+        dispatcher for all logical tasks mapped to that slot.
+
+State
+        A dense integer assigned to one logical task within a slot.  Codegen emits
```
luxfg, line 19: What is a "dense" integer?

spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+``reachable[A]``
+        The set of tasks reachable from ``A`` through outgoing inter-task edges.
+        If a trigger source is in ``reachable[A]``, then ``A`` must happen before
+        that source.
+
+``trigger_sources[B]``
+        The set of tasks that can directly trigger ``B``.  These are the immediate
+        predecessors of ``B`` in the task graph after DAG construction.
+
+With those relations, ``A`` is considered to safely precede ``B`` iff ``A``
```
luxfg: These definitions must be clarified:


``reachable[A]``
        The set of tasks reachable from ``A`` in the task graph. If ``C`` is in ``reachable[A]`` we say ``A`` is an _ancestor_ of ``C``. Note that every task is its own ancestor.

``trigger_sources[B]``
        The set of tasks that can directly trigger ``B``.  These are the immediate
        predecessors of ``B`` in the task graph after DAG construction.

With those relations, ``A`` is considered to _safely precede_ ``B`` iff ``A`` is an ancestor of every trigger source of ``B``.  
In code this is the predicate
implemented by :func:`_precedes_all_trigger_sources`.

Note that we must consider the trigger sources of ``B``, because a trigger source must be able to activate/unblock ``B``: For this, the task id of ``B`` must not be concurrently used by ``A``.

Two tasks conflict if neither direction holds:

* ``A`` does not safely precede all trigger sources of ``B``, and
* ``B`` does not safely precede all trigger sources of ``A``.

When that happens, the tasks may be simultaneously live from the point of view
of slot reuse and therefore need distinct hardware IDs.

spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+This module plans how logical CSL local tasks can share a smaller set of
+hardware local-task IDs when the program contains more local tasks than the
+target architecture exposes in :mod:`spatialstencil.syntax.csl.constants`.
+
+Terminology
+-----------
+
+Logical task
+        A ``tdag.CSLTask`` produced by ``create_csl_tasks``.  These are the tasks
+        the completion DAG wants to execute.
```
luxfg, line 11: The output of create_csl_tasks is not a "completion DAG" in the sense of the paper, but a "CSL Task graph" (AFTER fusion). That is, the input is assumed to consist of only ACTIVATE and UNBLOCK inter-task edges. In degree is at most 2.

luxfg: Then, the semantics of this CSL task graph is that an edge from ``A`` to ``B`` blocks execution of ``B`` before ``A`` completes.

spatialstencil/syntax/csl/task_recycling.py
luxfg: I confirmed the correctness of this condition.

spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+        return '\n'.join(lines) + '\n'
+
+
+def plan_task_bindings(
+    tasks: list[tdag.CSLTask],
+    task_creation_behavior: tdag.TaskCreationBehavior,
+    recycle_overflow_only: bool = True,
+) -> TaskBindingPlan:
+    """
+    Compute a local-task binding plan for the generated CSL.
```
luxfg: Add a comment that clarifies that tasks must be given in topological order.

spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+
+
+def _bounded_coloring(
+    conflict_graph: dict[int, set[int]],
+    topological_local_order: list[int],
+    max_colors: int,
+) -> dict[int, int] | None:
+    """Search for a coloring that uses at most ``max_colors`` colors.
+
+    The planner first tries greedy coloring.  If the greedy result exceeds the
```
luxfg: Using backtracking is exponential time in the number of vertices

I think its Theta(c^n) time

if no coloring exists, it may enumerate all possible colorings and actually achieve this exponential worst case.

spatialstencil/syntax/csl/task_recycling.py
```diff
@@ -0,0 +1,640 @@
...
+def _bounded_coloring_with_fixed_prefix(
+    conflict_graph: dict[int, set[int]],
+    topological_local_order: list[int],
+    max_colors: int,
+    fixed_coloring: dict[int, int],
+    max_slot_size: int,
+) -> dict[int, int] | None:
+    """Color the overflow tail with a fixed prefix and bounded slot occupancy.
+
+    This is an exact search.  ``fixed_coloring`` pins the unique non-recycled
```
luxfg: Again, exponential time.

luxfg (changes requested):
https://github.com/glukas/spatialStencil/pull/49

luxfg (approved):

PR #46, opened 2026-04-06
Support capturing relay DSD operations

tbennun:
Supports DSD generation of ops of the form `receive->operator->send`.
Also fixes a pass ordering issue where pruned fields would still appear in generated CSL code.


luxfg (approved):
Very nice, targeted PR!

PR #56, opened 2026-04-07
Limit local task bindings to not overlap with colors

tbennun:
Uses routed colors to limit task ID choice. On potential overlap that cannot be resolved, task recycling will take place.


luxfg (approved):
Good fix!!

PR #57, opened 2026-04-08
Revert phase inlining change and test stencil lowering

tbennun:
* Reverts unique stream names during phase inlining, which caused generating too many channels
* `test_benchmarks.sh` now regenerates the .sptl files every time
* Stream indentation has been corrected (generated .sptl files have no diff with repository)


tbennun:
@glukas clearly this is not the right solution - if channel is `auto`, the contract should be that the CSL code would not create conflicting channels. For some reason when lowering stencil IR to SPaDA we get many colors, even though the channel should be the same across phases.

luxfg:
Yeah, this needs further thought.

PR #67, opened 2026-08-17
Bounded streams

tbennun:
Adds stream bounds, the notion of stream epochs, and the `stream.close()` operation. This allows SpaDA to make use of Switches on the WSE.


irspec/docs/spatial/routing.md
```diff
@@ -167,3 +191,82 @@ to receive.
...
     If multicasting is used, the correctness conditions must be adapted accordingly, 
     especially when considering multiple phases.
+
+
+## Lowering to Switches
+
+!!! note "Note: Scope"
+    Everything in this section describes how the **Cerebras WSE / CSL backend** realizes epochs, not
+    the semantics of the Spatial IR itself. Colors, routers, switch positions and control messages
+    are properties of that target; a different backend may implement channel reuse by any means that
```
luxfg: Move these and following discussion to separate routing_wse document

irspec/docs/spatial/spatial.md
```diff
@@ -814,6 +822,75 @@ completion completion_name = foreach type k, type x in [0:K], receive(stream_nam
...
+    the close in [local order](../async#local-order). A use that is not ordered with respect to a
+    `close` on another PE is undefined behavior.*
+
+!!! danger "Error: Unclosed Stream"
+    Closing a stream is *collective*: every PE that sends on or receives from a stream must close
+    it. *Failing to close a stream on one of its participants raises a compile error where
+    detectable.*
+
+A stream that carries a [bound](#streams) closes itself once `BOUND` elements have been transferred;
+an explicit `close` on a bounded stream is redundant but legal. An unbounded stream is only closed
```
luxfg, line 856: Why allow this redundant close? does this crate timing questions? ie when is it “really” closed? Like if I await a redundant close, on does it brhave like a noop or like it closes there?

irspec/docs/spatial/spatial.md
```diff
@@ -814,6 +822,75 @@ completion completion_name = foreach type k, type x in [0:K], receive(stream_nam
...
+by an explicit `close`.
+
+!!! note "Note: Verification of Bounds"
+    Where the number of elements transferred over a bounded stream can be determined statically, it
```
luxfg: this is poorly defined. We do not define what “can be statically inferred means” and this is a shifting target. Rather say “when the compiler can infer the bounds statically, it MAY raise an exception”

luxfg (changes requested):
Very nice work!
Can you clarify some points in the spec & slightly restructure?

luxfg:
The 1d scalar was marked as “expected to fail” because of a know limitation on size-1 arrays, which should probably be fixed in a separate pr?

tbennun:
> The 1d scalar was marked as “expected to fail” because of a know limitation on size-1 arrays, which should probably be fixed in a separate pr?

there is a fix in progress, might make it to this pr because it's short

irspec/docs/spatial/spatial.md
tbennun: Fixed. Please check again.

irspec/docs/spatial/spatial.md
tbennun: The redundant close is allowed because the compiler automatically injects `close()` for every stream with an open epoch at the end of a phase. It was made for simplicity and will generate a no-op. We can change that with a bit more effort.

irspec/docs/spatial/routing.md
tbennun: Done.

luxfg (approved):
OK!

PR #70, opened 2026-08-25
Implement Cerebras Appliance Mode compiler and runtime entry points

### LINKED ISSUE #62: Integration of appliance mode compiler
In constrast to the basic mode, [(appliance mode)](https://sdk.cerebras.net/appliance-mode) provides no cslc compiler. Instead, one compiles directly in python by importing the compiler as class and calling its `compile()` method with appropriate arguments. My questions are:

- How can this compile workflow be implemented in SPADA?
- Should there be a cli flag for the user to set in order to activate appliance mode, or should SPADA try to guess appliance mode based on available cerebras python libs?
- Should the compiler call be directly imported into and implemented in  `spada/cli/compiler.py`, or should there be a wrapper script provided that acts like cslc in normal mode and that can be called without (major) changes from `spada/cli/compiler.py`?

tbennun:
On [Wafer-Scale Clusters](https://sdk.cerebras.ai/appliance-mode), compilation and runtime needs to run through appliance mode, which requires a different entry point to the compiler and runtime.

This new PR compiles with the SDK compiler, stages the appropriate files with SdkLauncher, and uses that launcher to run the script on the cluster directly.

Closes #62 as complete.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,11 @@ setup(
         ],
     },
     entry_points={
-        "console_scripts": ["sptlc=spada.cli.compiler:compile_spatial_ir"],
+        "console_scripts": [
+            "sptlc=spada.cli.compiler:compile_spatial_ir",
+            "sptlc-appliance=spada.cli.appliance_compiler:compile_spatial_ir_appliance",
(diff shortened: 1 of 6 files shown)


luxfg (approved):
OK!

PR #72, opened 2026-08-26
Vectorize nested multiply-accumulate loops into @fmac* DSD operations

### LINKED ISSUE #69: Nested multiply-accumulate loops lower to scalar code; targeted `@fmacs` vectorization gives 4.5–7.7× on `samples/spatial/blas/gemv.sptl`
## Summary

Compute blocks of the form

```
for i16 k, i16 l in [0:K, 0:K] {
    z[k] = z[k] + A_flat[k*K + l] * x[l]
}
```

(as written in `samples/spatial/blas/gemv.sptl`, phase 4) currently lower to scalar
nested `for (@range(...))` loops. The inner product therefore runs one element per
instruction, and its cost grows as K² while a hand-written CSL kernel using
`@fmacs` over strided DSDs stays nearly flat. We measured the crossover directly:
at K=16 the generated gemv becomes ~2× slower than a hand-written reference
(`gemv-collectives_2d` from Cerebras' sdk-examples), and ~4.6× slower at K=32.

We wrote a small, conservative lowering that turns this exact pattern into the
hand-written idiom and would like to contribute it, if there is interest.

## Why a compiler change (and not a source-level rewrite)

We first tried to express the vectorizable form at the language level and found
both routes closed:

- `await map i16 k in [0:K:1] { z[k] = z[k] + A_flat[(k*K) + l] * x[l] }` is
  rejected by `_dsd_from_array` ("Expected one index variable in array access") —
  the second loop variable `l` cannot appear in the access.
- A slice assignment `z[0:K:1] = z[0:K:1] + A_flat[l:K*K:K] * x[l]` is rejected
  by the surface grammar (`UnexpectedCharacters: No terminal matches ':'`),
  although `ArraySlice` supports strides at the IR level.

So users cannot opt into the fast form from `.sptl` today.

## Proposed lowering

In `emit_for`, detect the shape

```
for (k, l) in [0:Kk, 0:Kl]:  Z[k] = Z[k] + A[k*Ck + l*Cl + C0] * X[f(l)]
```

and emit the same idiom hand-written kernels use (cf. `gemv-collectives_2d`):

```
const __vecmac_dst = @get_dsd(mem1d_dsd, .{ .tensor_access = |i|{Kk} -> Z[i] });
const __vecmac_src = @get_dsd(mem1d_dsd, .{ .tensor_access = |i|{Kk} -> A[i * Ck] });
for (@range(i16, 0, Kl, 1)) |l| {
    const __vecmac_a = @increment_dsd_offset(__vecmac_src, l * Cl, f32);
    @fmacs(__vecmac_dst, __vecmac_dst, __vecmac_a, X[f(l)]);
}
```

Safety/conservatism:

- fires only on the exact affine MAC shape above, f32 only, ranges starting at 0
  with step 1; anything else falls back to the existing scalar loops;
- aliasing guard: not applied when `Z` is also read as `A` or `X` (the DSD op
  reorders reads relative to the sequential scalar loop);
- respects `--disable-dsd`, plus a new `--disable-mac-vectorization` flag wired
  the same way as the existing disable flags;
- `@increment_dsd_offset` and `@fmacs` are both available in the SDK line this
  repo targets: `@fmacs` is already emitted by `dsd_ops.py`, and
  `@increment_dsd_offset` is used by Cerebras' own `csl-examples` at tag
  `v1.4.0` (`benchmarks/cholesky/pe.csl`, lines 113–119).

## Measurements

Setup: WSE-2 `simfabric` (cycle counts are deterministic; every run verified
numerically against a NumPy reference), 4×4 PE grid, `gemv.sptl` with
`PX=PY=4`, fp32. Baseline is the unmodified scalar lowering; "hand-written" is
Cerebras' `gemv-collectives_2d` benchmark instrumented with the same
`<time>` timestamps the SPADA runtime uses. Kernel-window cycles
(`max(stop) − min(start)` across PEs):

| K | scalar lowering | vectorized | hand-written | speedup (vec vs scalar) |
|---|---|---|---|---|
| 4 | 490 | 579 | 3,127 | 0.85× |
| 8 | 1,521 | 809 | 3,378 | 1.88× |
| 16 | 8,357 | 1,851 | 4,090 | 4.51× |
| 32 | 28,955 | 3,744 | 6,274 | 7.73× |

The scalar lowering crosses above the hand-written kernel at K=16 and is 4.6×
slower by K=32; the vectorized lowering stays below it at every size we
measured. At K=4 the DSD setup constant makes vectorization slightly slower
(0.85×) — if that matters, a size threshold could gate the transform, but we
left the behavior unconditional and are flagging it here instead.

Caveats we want to be upfront about: measurements were taken on the SDK 2.10
simulator (with a small local compatibility layer for 2.10's CSL dialect; the
patch itself is written against current `main` and does not include that
layer). On unmodified `main` we verified that the transform fires and the
emitted code retains the SDK 1.4 dialect, and that the full
`tests/spatial_ir` + `tests/stencil_ir` suites pass (323 passed / 12 skipped),
plus 6 new unit tests covering the positive case, both aliasing cases, a
non-affine index, a mismatched accumulator, and the disable flag. We have not
run an end-to-end SDK 1.4 simulation.

## Offer

The change is ~180 lines of compiler code plus ~110 lines of tests
(`syntax/csl/statements.py` matcher + emitter, flag plumbing in
`lowering/spatial_ir_to_csl.py` and `cli/compiler.py`,
`tests/spatial_ir/test_mac_vectorization.py`). Happy to open a PR if this is
welcome, and to adjust scope (e.g., f16 via `@fmach`, a size threshold, or
folding the flag into an existing one) to your preference.

ycmath:
Implements the targeted MAC-loop vectorization from #69, incorporating the staging agreed there: this PR delivers the base vectorization together with asks (2), (3), and (5); asks (1) and (4) are deferred to a follow-up PR pending the `increment_dsd` semantics question raised in the issue thread.

## What this does

`emit_for` now recognizes the two-variable nested multiply-accumulate pattern

```
for (k, l) in [0:K, 0:L]:
    Z[k] = Z[k] + A[k*Ck + l*Cl + C0] * X[f(l)]
```

and lowers it to a strided base DSD over `A` plus a per-`l` `@increment_dsd_offset` and one `@fmac*` per column, instead of scalar loops — the same idiom as the handwritten CSL GEMV. On `samples/spatial/blas/gemv.sptl` this measured 4.5–7.7× (details in #69).

- **Loop order (ask 2):** operand roles are derived from the accumulator index — the loop variable indexing `Z` with coefficient 1 takes the `k` role, the other is the reduction variable — so `for (l, k)` vectorizes identically to `for (k, l)`. Per-element accumulation order over `l` is preserved bitwise in both orders.
- **Type generality (ask 3):** dispatch mirrors `FMADSDOp._as_csl`: all-f16 → `@fmach`, f32 accumulate with f16 scalar → `@fmachs`, all-f32 → `@fmacs`; the `@increment_dsd_offset` element type follows `A`. Unsupported combinations fall back to scalar loops.
- **No separate flag (ask 5):** the vectorizer sits behind `--disable-dsd` like all other DSD detection.

Conservative by construction — falls back to scalar code on: aliasing of `Z` with `A` or `X` (DSD ops reorder reads relative to the sequential loop), non-affine indices, non-unit steps, nonzero `k` start, or any dtype combination outside the table above.

## Tests

`tests/spatial_ir/test_mac_vectorization.py`: 11 tests covering the positive matrix (both loop orders × f32/f16/mixed dtypes) and negative cases (aliasing in both orders, non-affine index, accumulator mismatch, unsupported dtypes, `--disable-dsd`). Full suite: 386 passed, 12 skipped on this branch.

`@increment_dsd_offset` is available from SDK 1.x (Cerebras' own csl-examples v1.4.0 cholesky uses it), so generated code stays compatible with the SDK 1.4 pinned in CI; the f16 form of the type parameter was additionally verified to compile with cslc.

### DIFF
--- a/spada/syntax/csl/statements.py
+++ b/spada/syntax/csl/statements.py
@@ -250,10 +250,253 @@ def emit_assignment(statement: spir.AssignmentStatement, dsds: UniqueDSDDict, dt
     return dsd_ops.DSD_ASSIGNMENT_MAPPING[dsd_op]()
 
 
+# Targeted vectorization of the nested multiply-accumulate loop
+#   for (k, l) or (l, k) in either order: Z[k] = Z[k] + A[k*Ck + l*Cl + C0] * X[f(l)]
+# into the handwritten CSL idiom (strided base DSD + per-l @increment_dsd_offset + @fmac*).
+# The accumulator index decides the roles: the loop variable indexing Z is k, the other
+# is the reduction variable l. Dtypes dispatch like FMADSDOp._as_csl (@fmach / @fmachs /
+# @fmacs). Conservative: fires only when Z is distinct from A and X; anything else falls
+# through to scalar loops. Gated behind --disable-dsd like all other DSD detection.
+# @increment_dsd_offset is available from SDK 1.x (used by Cerebras' own csl-examples
+# v1.4.0 cholesky benchmark).
+
+
+def _affine_of(expr: spir.Expression, varnames: set[str]) -> Optional[tuple[dict[str, int], int]]:
+    """
+    Decomposes an index expression into an affine form over the given loop variables.
+
+    :param expr: The index expression to decompose.
+    :param varnames: The loop-variable names that may carry coefficients.
+    :return: A ``(coefficients, constant)`` pair such that the expression equals
+             ``sum(coefficients[var] * var) + constant`` with integer coefficients,
+             or None if the expression is not affine in the given variables.
+    """
+    e = expr.value if isinstance(expr, spir.Expression) else expr
+    if isinstance(e, spir.Identifier):
+        if e.name in varnames:
+            return {e.name: 1}, 0
+        return None
+    if isinstance(e, (spir.ConstantLiteral, spir.Parameter)):
+        try:
+            v = e.eval()
+        except (ValueError, TypeError):
+            return None
+        return ({}, int(v)) if isinstance(v, int) else None
+    if isinstance(e, spir.UnaryOperator) and e.op == '-':
+        sub = _affine_of(e.value, varnames)
+        if sub is None:
+            return None
+        return {k: -c for k, c in sub[0].items()}, -sub[1]
+    if isinstance(e, spir.BinaryOperator):
+        left = _affine_of(e.left, varnames)
+        right = _affine_of(e.right, varnames)
+        if e.op == '+' and left and right:
+            coeffs = dict(left[0])
+            for k, c in right[0].items():
+                coeffs[k] = coeffs.get(k, 0) + c
+            return coeffs, left[1] + right[1]
+        if e.op == '-' and left and right:
+            coeffs = dict(left[0])
+            for k, c in right[0].items():
+                coeffs[k] = coeffs.get(k, 0) - c
+            return coeffs, left[1] - right[1]
+        if e.op == '*' and left and right:
+            if not left[0]:      # left operand is a pure constant
+                s = left[1]
+                return {k: c * s for k, c in right[0].items()}, right[1] * s
+            if not right[0]:     # right operand is a pure constant
+                s = right[1]
+                return {k: c * s for k, c in left[0].items()}, left[1] * s
+        return None
+    return None
+
+
+def _ids_in(expr: spir.Expression) -> set[str]:
+    """
+    Collects the names of all identifiers appearing in an expression.
+
+    :param expr: The expression to walk.
+    :return: The set of identifier names.
+    """
+    e = expr.value if isinstance(expr, spir.Expression) else expr
+    return {n.name for n in e.walk() if isinstance(n, spir.Identifier)}
+
+
+def _single_index(node: spir.SpatialNode) -> Optional[spir.Expression]:
+    """
+    Extracts the index of a one-dimensional single-element array access.
+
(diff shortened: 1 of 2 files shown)


ycmath:
Note for review: #75 was an accidental duplicate of this PR from a parallel workflow on my side and is now closed — this PR is the canonical one.

It is based on 7be46fb but merges cleanly onto current `main` (post-#70); on the merged tree the full suite is 386 passed / 12 skipped. On that merged tree the generated CSL also compiles with `cslc` (SDK 2.10) — both the `samples/spatial/blas/gemv.sptl` output and a minimal mixed-dtype kernel whose `@fmachs` form (f32 accumulator and DSD operand, f16 scalar) is the one this PR emits. Note that the branch needs the merge to compile under SDK 2.10 at all: its pre-#70 base still emits `param memcpy_params: comptime_struct;`, which 2.10 rejects independently of this change.

One `@fmachs` form is still left on the table: the SDK's own `csl-libs/dsd_ops.csl` also permits an f16 DSD operand with an f32 accumulator, which this pass declines and lowers to a scalar loop. Happy to add that (plus integer-dtype fallback tests) here or in the follow-up, whichever you prefer.

PR #73, opened 2026-08-26
Fix non-memcpy benchmark path saving raw timestamps instead of cycle counts

ycmath:
In the non-memcpy benchmark path, `Program.run` calls `copy_back_benchmark_data`, which returns a `(cycle_start, cycle_stop)` tuple of absolute timestamps, and binds the tuple to `cycle_counts`. Consequently:

- `perf_cycles.npy` stores the two stacked timestamp arrays instead of elapsed cycle counts, and
- the printed "Cycle count" summary is computed from absolute timestamps.

The memcpy path already computes the difference correctly. This changes the call to `copy_back_benchmark_cycles`, which returns `cycle_stop - cycle_start`, making both paths consistent.

### DIFF
--- a/spada/runtime/runtime.py
+++ b/spada/runtime/runtime.py
@@ -389,7 +389,7 @@ class Program:
             print("Copy-back complete.", flush=True)
 
             if self.benchmark and not self.metadata.memcpy_mode:
-                cycle_counts = copy_back_benchmark_data(self.runtime, self.metadata)
+                cycle_counts = copy_back_benchmark_cycles(self.runtime, self.metadata)
                 np.save(self.output_dir / "perf_cycles.npy", cycle_counts)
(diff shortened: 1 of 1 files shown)


tbennun (approved):
LGTM!

ycmath:
Thanks for the review! For the record, the `test-csl` failure here is a fork-PR limitation rather than this change: the job fetches the SDK with `wget "${{ secrets.CSL_SDK_URL }}"`, and GitHub does not pass repository secrets to workflows triggered by PRs from forks, so the URL comes through empty (log: `wget -q -O sdk-1.4.0.tar.gz ""` → exit 1; the `Linux-deps` cache had also expired). `test-pytest` passes, and `main` itself is green. Same constraint applies to #72.

PR #75, opened 2026-09-01
Generalize MAC vectorization: either loop order, f16/f32/mixed dtypes, fold into --disable-dsd

ycmath:
Follow-up to #69 — first of the two stages proposed there. This covers asks (2), (3), and (5); arbitrary nesting depth (1) and generalizing the affine approach to any DSD operation (4) are left for a second PR, pending the `increment_dsd` relative-vs-absolute offset question in the issue thread.

**Loop order (2).** Loop-variable roles are now detected from the destination index's affine decomposition — the variable appearing as the bare destination index is the row variable, the other is the reduction variable — so both `for (k, l)` and `for (l, k)` declaration orders vectorize. The positional hardcoding is gone.

**Dtypes (3).** The f32 gate is replaced by a dispatch that mirrors `FMADSDOp._as_csl` exactly (sharing `dsd_ops._get_base_dtype`): `@fmach`, `@fmachs`, `@fmacs` for the same dtype combinations FMADSDOp models. Integer dtypes are intentionally out of scope — `DSD_ASSIGNMENT_MAPPING` carries no integer MAC builtin. Per the SDK language reference, `@increment_dsd_offset`'s `elem_type` accepts any ABI-compatible numeric type including `f16`, so the offset-stepping side generalizes with it.

**Flag (5).** `--disable-mac-vectorization` is removed; the pass is now gated solely by `--disable-dsd`, as suggested. This changes nothing for existing `--disable-dsd` users (the pass was already AND-gated behind it). The negative test that isolated MAC vectorization from other DSD emission was rewritten against `--disable-dsd`.

**Tests.** `test_mac_vectorization.py` grows 6 → 20: positives across order × dtype, negatives for unsupported dtypes, and the existing aliasing / non-affine / accumulator-mismatch negatives replicated across both loop orders. Full suite: 395 passed / 12 skipped (baseline at eaf8d1c: 375 / 12 — zero regressions). f32 gemv codegen emits a byte-identical vectorized MAC block relative to the pre-change output.

### DIFF
--- a/spada/cli/appliance_compiler.py
+++ b/spada/cli/appliance_compiler.py
@@ -92,6 +92,7 @@ def compile_spatial_ir_appliance(input_file: str, output_folder: str, param: lis
                                  generate_only: bool, simulator: bool, mgmt_namespace: Optional[str],
                                  resource_cpu: Optional[int], resource_mem: Optional[int], check_version: bool,
                                  disable_benchmarking: bool, disable_asynchronous: bool, disable_dsd: bool,
+                                 disable_mac_vectorization: bool,
                                  disable_map: bool, disable_task_fusion: bool, disable_task_recycling: bool,
                                  disable_copy_elision: bool, disable_close_elision: bool, disable_switching: bool):
     program = generate_program(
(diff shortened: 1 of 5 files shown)


ycmath:
Closing as a duplicate: #72 (opened earlier from a parallel workflow on my side) covers the same asks (2), (3), and (5). Sorry for the noise — #72 is the canonical one.

